Issue #3, opened 2015-03-03
Add support for services for Connext API


wjwwood:
Acceptance Criteria:

- [x] Make use of ros_middleware_interface service API

esteve:
@dirk-thomas @tfoote @wjwwood this ticket depends on a bunch of PRs, when you have a moment, I'd appreciate feedback for the remaining ones.

Thanks.

dirk-thomas:
Beside the two small comments they LGTM.

wjwwood:
I just reviewed the four pr's, lgtm +1

PR #4, opened 2015-03-06
Add support for services in rmw_connext

esteve:
This PR adds support for service in `rmw_connext`. The big changes are:

- Use the C API and structures
- Updates to `rmw_take*` functions caused by ros2/rmw#4

Connects to ros2/rmw_connext#3

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -1,3 +1,5 @@
+#include <rmw/types.h>
+
 #include <iostream>
 #include <stdexcept>
 
@@ -40,7 +42,7 @@ void init()
     };
 }
 
-ros_middleware_interface::NodeHandle create_node()
+rmw_node_t * create_node(const char * name)
 {
     std::cout << "create_node()" << std::endl;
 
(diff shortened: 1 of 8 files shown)


rmw_connext_cpp/src/functions.cpp
```diff
@@ -1,3 +1,5 @@
+#include <rmw/types.h>
```
dirk-thomas: We should move this include down to be grouped with the other ones from the same package.

esteve: Fixed, GitHub is too smart and doesn't detect that this comment is outdated, though :-P

rmw_connext_cpp/src/functions.cpp
```diff
@@ -169,48 +167,41 @@ ros_middleware_interface::PublisherHandle create_publisher(const ros_middleware_
...
-    //std::cout << "publish()" << std::endl;
-
-    if (publisher_handle._implementation_identifier != _rti_connext_identifier)
+    if (publisher->implementation_identifier != _rti_connext_identifier)
     {
-        printf("publisher handle not from this implementation\n");
-        printf("but from: %s\n", publisher_handle._implementation_identifier);
-        throw std::runtime_error("publisher handle not from this implementation");
+        rmw_set_error_string("publisher handle not from this implementation");
+        // rmw_set_error_string("but from: %s\n", publisher->implementation_identifier);
```
tfoote: Is this not possible anymore?

tfoote:
It looks ok to me. Though there appears to be quite a bit of info lost which used to be embedded in the error messages.

rmw_connext_cpp/src/functions.cpp
esteve: `rmw_set_error_string` does not accept format strings, but it'd be a nice addition IMO.

PR #7, opened 2015-03-11
fix check of found connext libraries

dirk-thomas:
This checks explicitly for the expected libraries. Since 600b4bf97ed7e6e5b3db081d7a01a691d0a27d7f the previous checks were not sufficient anymore.

### DIFF
--- a/connext_cmake_module/cmake/Modules/FindConnext.cmake
+++ b/connext_cmake_module/cmake/Modules/FindConnext.cmake
@@ -27,19 +27,47 @@
 
 set(Connext_FOUND FALSE)
 
+# check if all libraries with an expected name have been found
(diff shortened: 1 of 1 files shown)


esteve:
+1

tfoote:
+1

dirk-thomas:
If you have a symlink for "rtiddsgen2" point from `~/RTI/bin` to the binary form the Debian package you will need to recreate it since the update CMake module will report the actual location: `lib/PLATFORM-SPECIFIC/rtiddsgen2`.

Since the script in that place is not working out of the box you need to replace it with a symlink pointing to the working file from the Debian package.

wjwwood:
+1

PR #8, opened 2015-03-12
Windows support

wjwwood:
This pull request fixes Windows related issues which prevent all of ROS 2 from compiling.

This is a work in progress.

Connects to ros2/ros2#9

### DIFF
--- a/connext_cmake_module/CMakeLists.txt
+++ b/connext_cmake_module/CMakeLists.txt
@@ -1,8 +1,20 @@
 cmake_minimum_required(VERSION 2.8.3)
 
-project(connext_cmake_module NONE)
+project(connext_cmake_module)
 
 find_package(ament_cmake REQUIRED)
+
+list(INSERT CMAKE_MODULE_PATH 0 "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
+find_package(Connext MODULE)
+if(Connext_HOME)
+  normalize_path(Connext_HOME "${Connext_HOME}")
+  normalize_path(Connext_LIBRARY_DIR "${Connext_LIBRARY_DIR}")
+  if(WIN32)
+    ament_environment_hooks(env_hook/connext.bat.in)
+  else()
+    ament_environment_hooks(env_hook/connext.sh.in)
+  endif()
+endif()
 
 if(AMENT_ENABLE_TESTING)
   find_package(ament_lint_auto REQUIRED)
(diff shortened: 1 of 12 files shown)


tfoote:
lgtm

esteve:
+1

connext_cmake_module/CMakeLists.txt
```diff
@@ -1,9 +1,21 @@
 cmake_minimum_required(VERSION 2.8.3)
 
-project(connext_cmake_module NONE)
+project(connext_cmake_module)
```
dirk-thomas: Why remove `NONE` here? The package does seem to compile anything.

rmw_connext_xtypes_dynamic_cpp/CMakeLists.txt
```diff
@@ -24,7 +26,7 @@ ament_export_dependencies(rmw rosidl_generator_cpp)
 
 register_rmw_implementation()
 
-link_directories(${Connext_LIBRARY_DIRS})
+link_directories("${Connext_LIBRARY_DIRS}")
```
dirk-thomas: Why does this need quotes?

wjwwood: If one of the paths contains whitespace they will be interpreted as two paths. E.g. a value of `C:\foo;C:\Program Files\bar;C:\baz` would be considered `link_directories("C:\foo;C:\Program" "Files\bar;C:\baz")` which would produce this when iterated on: `C:\foo`, `C:\Program`, `Files\bar`, and `C:\baz`.

At least that's what I observed, maybe there is a better way to fix it.

connext_cmake_module/CMakeLists.txt
wjwwood: If you put `NONE` there then this variable `CMAKE_SIZEOF_VOID_P` is not set and you cannot include the `FindConnext.cmake` file in order to fill out the env hook.

dirk-thomas: It looks like `FindConnext.cmake` should make sure to perform the necessary compile checks in order to be used from this with `NONE` being set.

rmw_connext_xtypes_dynamic_cpp/CMakeLists.txt
dirk-thomas: Sounds reasonable. I guess that might apply to other locations too. Will have to try this once we have the CI job running, e.g. always build in a location with spaces (@tfoote fyi).

connext_cmake_module/CMakeLists.txt
dirk-thomas: I thought `enable_language` would be the way to go but the docs suggest that it doesn't set the desired variables. Just adding a comment here why we can't use `NONE` should be good for now.

PR #10, opened 2015-03-19
Use identifiers as symbols to uniquely identify these rmw implementations

esteve:
Connects to ros2/ros2#18

### DIFF
--- a/rmw_connext_xtypes_dynamic_cpp/CMakeLists.txt
+++ b/rmw_connext_xtypes_dynamic_cpp/CMakeLists.txt
@@ -19,7 +19,7 @@ find_package(rosidl_generator_cpp REQUIRED)
 
 find_package(rosidl_typesupport_introspection_cpp REQUIRED)
 ament_export_dependencies(rosidl_typesupport_introspection_cpp)
-
+ament_export_include_directories(include/${PROJECT_NAME}/impl ${Connext_INCLUDE_DIRS})
 ament_export_dependencies(rmw rosidl_generator_cpp)
 
 register_rmw_implementation()
@@ -31,6 +31,7 @@ ament_target_dependencies(rmw_connext_xtypes_dynamic_cpp
   "rmw"
   "rosidl_generator_cpp"
   "Connext")
+configure_rmw_library(rmw_connext_xtypes_dynamic_cpp)
 
 ament_export_libraries(rmw_connext_xtypes_dynamic_cpp dl pthread)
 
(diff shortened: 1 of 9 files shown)


esteve:
@dirk-thomas @tfoote @wjwwood ping

rmw_connext_xtypes_dynamic_cpp/include/rmw_connext_xtypes_dynamic_cpp/impl/rmw/get_type_support_handle.h
```diff
@@ -0,0 +1,23 @@
...
+extern "C" {
+  extern const char * rti_connext_dynamic_identifier;
+}
+
+namespace rmw
+{
+template<typename T>
+const rosidl_message_type_support_t * get_type_support_handle()
+{
+  std::cout << rti_connext_dynamic_identifier << std::endl;
```
dirk-thomas: Remove debug output?

rosidl_typesupport_connext_cpp/cmake/rosidl_typesupport_connext_cpp_generate_interfaces.cmake
```diff
@@ -129,5 +129,6 @@ install(
   DESTINATION "lib"
 )
 
-ament_export_include_directories(include)
 ament_export_libraries(${rosidl_generate_interfaces_TARGET}${_target_suffix} ${Connext_LIBRARIES})
+
+ament_export_include_directories(include)
```
dirk-thomas: Why moving this line around? The previous order (include before libraries) seems to make sense.

rmw_connext_xtypes_dynamic_cpp/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ find_package(rosidl_generator_cpp REQUIRED)
 
 find_package(rosidl_typesupport_introspection_cpp REQUIRED)
 ament_export_dependencies(rosidl_typesupport_introspection_cpp)
-
+ament_export_include_directories(include/${PROJECT_NAME}/impl ${Connext_INCLUDE_DIRS})
```
dirk-thomas: I would suggest moving the include dir export either above or below both dependency exports.

rosidl_typesupport_connext_cpp/cmake/rosidl_typesupport_connext_cpp_generate_interfaces.cmake
esteve: To make it more consistent with `rmw_opensplice`. I also was going to add support for "native" `Duration_t` and `Time_t`, so it needs to skip most of the CMake file if the package name is `builtin_msgs`

dirk-thomas: Makes sense.

tfoote:
lgtm

dirk-thomas:
lgtm

PR #11, opened 2015-03-25
Fix get type support handle

wjwwood:
This pull request updates the layout of the connext packages to work with https://github.com/ros2/rosidl/pull/9 as well as https://github.com/ros2/rosidl_dds/pull/9 in the case of `rmw_connext_xtypes_dynamic_cpp`. Additionally, I fixed some stuff related to using Connext on OS X, fixed linking on Windows, and I think I resolved the issue you guys were having running `rtiddsgen2` on Linux.

Connects to ros2/ros2#26


rosidl_typesupport_connext_cpp/include/rosidl_typesupport_connext_cpp/impl/rosidl_generator_cpp/get_type_support_handle_impl.hpp
```diff
@@ -0,0 +1,42 @@
...
+const rosidl_message_type_support_t * get_type_support_handle()
+{
+  // Hand off to implementation specific handle getter.
+  // By using this implementation specific getter in this header, it makes
+  // the user executable or library have a reference to an implementation
+  // specific symbol, which is implemented in the implementation specific
+  // message library. This is intentional to allow the linker to pick the
+  // correct implementation library when being over linked with multiple
+  // implementation options.
+  return rmw_connext_cpp::get_type_support_handle_impl<T>();
```
esteve: Would renaming `rmw_connext_cpp::get_type_support_handle_impl<T>();` to `rmw_connext_cpp::get_type_support_handle_connext_cpp_impl<T>();` work? It'd make searching easier and remove ambiguity.

tfoote:
+1

rosidl_typesupport_connext_cpp/include/rosidl_typesupport_connext_cpp/impl/rosidl_generator_cpp/get_service_type_support_handle.h
```diff
@@ -1,14 +1,14 @@
 #ifndef __rmw__get_service_type_support_handle__h__
 #define __rmw__get_service_type_support_handle__h__
 
-#include "rosidl_typesupport_connext_cpp/ServiceTypeSupport.h"
+#include "rosidl_typesupport_opensplice_cpp/ServiceTypeSupport.h"
```
dirk-thomas: `opensplice` looks wrong here.

rosidl_typesupport_connext_cpp/include/rosidl_typesupport_connext_cpp/impl/rosidl_generator_cpp/get_type_support_handle_impl.hpp
dirk-thomas: I don't think `rosidl_typesupport_connext_cpp` should refer to anything in the namespace `rmw_connext_cpp`. Currently there might be a one-to-one mapping but conceptionally a specific type support can be used by multiple rmw implementations. And this is actually the case for the introspection one e.g.

rosidl_typesupport_connext_cpp/include/rosidl_typesupport_connext_cpp/impl/rosidl_generator_cpp/get_service_type_support_handle.h
wjwwood: I reverted this particular change since the diff was so small anyways. (it was also moved, but I didn't change that. See: 4288605

rosidl_typesupport_connext_cpp/include/rosidl_typesupport_connext_cpp/impl/rosidl_generator_cpp/get_type_support_handle_impl.hpp
wjwwood: Yeah, we could do that since it is implementation specific, but I was trying to avoid the names getting long in the extreme, plus it is in an implementation specific namespace. I'm not opposed to changing it, what do others think?

wjwwood: @dirk-thomas I agree, I can update the namespace of that function to be `rosidl_typesupport_connext_cpp`.

wjwwood: I'll also do the same for OpenSplice.

wjwwood: Addressed in e9f1951 and https://github.com/ros2/rmw_opensplice/commit/303e1a6e7ef505a43533eec7ecc2bd8f53f502fd

rosidl_typesupport_connext_cpp/CMakeLists.txt
```diff
@@ -16,7 +16,10 @@ find_package(ament_cmake_python REQUIRED)
 
 ament_export_dependencies(rosidl_cmake)
 ament_export_dependencies(rosidl_generator_dds_idl)
-ament_export_include_directories(include)
+# The reason the impl folder is exported is that it contains the implementation
+# for the get_type_support_handle function and defines the opensplice specific
```
dirk-thomas: Copy-n-paste from "opensplice".

PR #12, opened 2015-03-30
[xtypes] publisher / subscriber segfaults

dirk-thomas:
Example to reproduce: userland unit test `test_(publisher|subscriber)__rmw_connext_xtypes_dynamic_cpp`

### DIFF
--- a/rmw_connext_xtypes_dynamic_cpp/src/functions.cpp
+++ b/rmw_connext_xtypes_dynamic_cpp/src/functions.cpp
@@ -202,7 +202,8 @@ rmw_create_publisher(const rmw_node_t * node,
 
     DDSDomainParticipant* participant = (DDSDomainParticipant*)node->data;
 
-    if (type_support->typesupport_identifier != rosidl_typesupport_introspection_cpp::typesupport_introspection_identifier)
+    // TODO avoid string comparison
+    if (strncmp(type_support->typesupport_identifier, rosidl_typesupport_introspection_cpp::typesupport_introspection_identifier, strlen(rosidl_typesupport_introspection_cpp::typesupport_introspection_identifier)))
     {
         rmw_set_error_string("type support not from this implementation");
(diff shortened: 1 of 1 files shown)


dirk-thomas:
This patch fixed the current segfault. But we have to revisit the type support identifiers and restore the checking functionality it should provide without requiring string comparison.

I have created a separate ticket since that will likely involve more effort: ros2/ros2#30

@esteve @tfoote @wjwwood Please review this workaround.

rmw_connext_xtypes_dynamic_cpp/src/functions.cpp
```diff
@@ -202,7 +202,8 @@ rmw_create_publisher(const rmw_node_t * node,
 
     DDSDomainParticipant* participant = (DDSDomainParticipant*)node->data;
 
-    if (type_support->typesupport_identifier != rosidl_typesupport_introspection_cpp::typesupport_introspection_identifier)
+    // TODO avoid string comparison
+    if (strcmp(type_support->typesupport_identifier, rosidl_typesupport_introspection_cpp::typesupport_introspection_identifier))
```
esteve: Could you replace this with `strncmp`?

dirk-thomas: What would you suggest to pass as `n`? Because if I use `strlen` to determine the length the problem would be same if the string wouldn't have a terminating zero character.

esteve: We know the length of the identifier (e.g. rosidl_typesupport_introspection_cpp::typesupport_introspection_identifier), we could just use that.

dirk-thomas: Updated to use `strncmp(A, B, strlen(B))`. Thanks.

esteve:
+1

wjwwood:
+1

PR #14, opened 2015-04-06
fix finding Connext

dirk-thomas:
This fixes finding Connext for me. Without the patch it find every library twice (one with two consecutive slashes). I am using the binary package of Connext - not the Debian package.

@esteve Please check if this still works for you with the Debian package.


esteve:
+1 Works with the Debian packages and `NDDSHOME` set to `/usr`

wjwwood:
+1

PR #17, opened 2015-04-10
Added support for non-primitive types

### LINKED ISSUE #16: unbounded fields workaround does not work for non-primitive datatypes in srvs
I created rcl_interfaces with some basic nested service definitions: https://github.com/ros2/rcl_interfaces/tree/master/rcl_interfaces/srv

However it crashed on generation: 
```
WARN com.rti.ndds.nddsgen.antlr.auto.IdlLexer ParamDescription_.idl line 33  preprocessor directive not supported. It will be ignored
WARN com.rti.ndds.nddsgen.antlr.auto.IdlLexer GetParamsRequest_.idl line 25  preprocessor directive not supported. It will be ignored
INFO com.rti.ndds.nddsgen.Main Done
Traceback (most recent call last):
  File "/home/tfoote/work/ros2/parameters/install/lib/rosidl_typesupport_connext_cpp/rosidl_typesupport_connext_cpp", line 75, in <module>
    sys.exit(main())
  File "/home/tfoote/work/ros2/parameters/install/lib/rosidl_typesupport_connext_cpp/rosidl_typesupport_connext_cpp", line 61, in main
    service_specs,
  File "/home/tfoote/work/ros2/parameters/install/lib/python3.4/site-packages/rosidl_typesupport_connext_cpp/__init__.py", line 86, in generate_dds_connext_cpp
    _modify(plugin_cxx_filename, unbounded_fields, _step_2_1_and_2_3_and_2_4)
  File "/home/tfoote/work/ros2/parameters/install/lib/python3.4/site-packages/rosidl_typesupport_connext_cpp/__init__.py", line 94, in _modify
    modified = callback(unbounded_fields, lines)
  File "/home/tfoote/work/ros2/parameters/install/lib/python3.4/site-packages/rosidl_typesupport_connext_cpp/__init__.py", line 144, in _step_2_1_and_2_3_and_2_4
    modified |= _step_2_3(unbounded_fields, lines)
  File "/home/tfoote/work/ros2/parameters/install/lib/python3.4/site-packages/rosidl_typesupport_connext_cpp/__init__.py", line 188, in _step_2_3
    dds_type = _get_dds_type(unbounded_fields, field_name)
  File "/home/tfoote/work/ros2/parameters/install/lib/python3.4/site-packages/rosidl_typesupport_connext_cpp/__init__.py", line 250, in _get_dds_type
    idl_type = MSG_TYPE_TO_IDL[field.type.type]
KeyError: 'ParamDescription'
```

With a little debugging I localized it to this workaround for unbounded arrays: https://github.com/ros2/rmw_connext/blob/master/rosidl_typesupport_connext_cpp/rosidl_typesupport_connext_cpp/__init__.py#L76-L86

It is using a strict dictionary of primitive types. It needs to be robust to unbounded non-primitive types. 

There is a more complete method msg_type_to_idl that would probably work, but the logic needs to be updated to support the arbitrary datatypes, not just the list of primitives.

esteve:
Fixes #16

### DIFF
--- a/rosidl_typesupport_connext_cpp/rosidl_typesupport_connext_cpp/__init__.py
+++ b/rosidl_typesupport_connext_cpp/rosidl_typesupport_connext_cpp/__init__.py
@@ -247,8 +247,11 @@ IDL_TYPE_TO_DDS = {
 
 def _get_dds_type(fields, field_name):
     field = [f for f in fields if '%s_' % f.name == field_name][0]
-    idl_type = MSG_TYPE_TO_IDL[field.type.type]
-    return IDL_TYPE_TO_DDS[idl_type]
+    if field.type.is_primitive_type():
+      idl_type = MSG_TYPE_TO_IDL[field.type.type]
+      return IDL_TYPE_TO_DDS[idl_type]
+    else:
+      return '%s::dds_::%s_' % (field.type.pkg_name, field.type.type)
 
 
 def generate_cpp(pkg_name, message_specs, service_specs, output_dir, template_dir):


wjwwood:
lgtm, obviously pending @tfoote's testing.

dirk-thomas:
This looks very similar to the logic in `msg_type_to_idl` in the `rosidl_generator_cpp_idl` package. It should probably reuse that logic (which will need refactoring to expose parts of the function).

esteve:
@dirk-thomas similar, but slightly different. This change is for generating the proper code for calling `Seq_set_maximum` on sequences of non primitive types. For parameters, this change generates the following code:

```cpp
rcl_interfaces::dds_::Param_Seq_set_maximum(&sample->parameters_, 0);  /* ROSIDL */
```

dirk-thomas:
The function (or part of it) does perform exactly the mapping you are creating here: https://github.com/ros2/rosidl_dds/blob/c7eddca2925df983eaa8e6ddc52d60386ea80893/rosidl_generator_dds_idl/rosidl_generator_dds_idl/__init__.py#L201-L204

And since the package `rosidl_generator_dds_idl` defines this mapping in the templates I think this package should ask `rosidl_generator_dds_idl` for doing the mapping rather then doing it again. E.g. if the templates in `rosidl_generator_dds_idl` would change the function would too, but your package would still rely on a custom implementation.

esteve:
@dirk-thomas this is the code that gets generated if that snippet is replaced with `msg_type_to_idl`, which is obviously invalid:

```cpp
['', '', 'sequence<rcl_interfaces::dds_::Param_>']Seq_set_maximum(&sample->parameters_, 0);  /* ROSIDL */
```

so I could add these two functions `rosidl_generator_dds_idl` that did this and modify `msg_type_to_idl`:

```python

def get_idl_type(type_):
    if type_.is_primitive_type():
        idl_type = MSG_TYPE_TO_IDL[type_.type]
    else:
        idl_type = '%s::dds_::%s_' % (type_.pkg_name, type_.type)

def get_dds_type(type_):
    idl_type = get_idl_type(type_)
    return IDL_TYPE_TO_DDS.get(idl_type, '%s::dds_::%s_' % (field.type.pkg_name, field.type.type))

def msg_type_to_idl(type_):
    idl_type = get_idl_type(type_)
    return _msg_type_to_idl(type_, idl_type)
```

however, I don't think all this de-duplication makes the code more readable. And if this is a matter of whether this is the correct approach because `rosidl_generator_dds_idl` is the place where this code should go, that's debatable because this is a workaround that's only used in `rmw_connext`

tfoote:
I've tested it and it builds fine. 

I agree that the current format is a little bit more readable, but for maintainability in the future I think that having things like the IDL_TYPE_TO_DDS lookup table and functions should be consolidated to rosidl_generator_dds_idl would be better. We currently only use them in for connext, but they are providing the same type of mappings and may be useful for other implementations in the future. And also if they're colocated it is easier to make sure they are consistent going forward.

dirk-thomas:
The merged code fails the PEP8 tests.

esteve:
@dirk-thomas the code didn't pass PEP8 before, but the code that was merged does, that's a slight difference :-) I'll have a look at it anyway.

esteve:
@dirk-thomas I stand corrected, the code has wrong indentation, oops! :-)

esteve:
PEP8 issues fixed in master.

PR #24, opened 2015-04-28
export type support for rmw implementation

dirk-thomas:
This is necessary to build libraries / executable against messages / services within the same package.

Connects to ros2/rmw#11.


wjwwood:
+1

PR #27, opened 2015-04-29
implement services for dynamic data

dirk-thomas:
Connects to ros2/ros2#27

### DIFF
--- a/rmw_connext_xtypes_dynamic_cpp/src/functions.cpp
+++ b/rmw_connext_xtypes_dynamic_cpp/src/functions.cpp
@@ -18,7 +18,10 @@
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wall"
 #pragma GCC diagnostic ignored "-Wdeprecated-register"
+#include <ndds/connext_cpp/connext_cpp_replier_details.h>
+#include <ndds/connext_cpp/connext_cpp_requester_details.h>
 #include <ndds/ndds_cpp.h>
+#include <ndds/ndds_requestreply_cpp.h>
 #pragma GCC diagnostic pop
 
 #include <rmw/allocators.h>
@@ -32,6 +35,7 @@
 #include "rosidl_typesupport_introspection_cpp/field_types.hpp"
 #include "rosidl_typesupport_introspection_cpp/identifier.hpp"
 #include "rosidl_typesupport_introspection_cpp/message_introspection.hpp"
+#include "rosidl_typesupport_introspection_cpp/service_introspection.hpp"
 
 // This extern "C" prevents accidental overloading of functions. With this in
 // place, overloading produces an error rather than a new C++ symbol.
@@ -326,7 +330,7 @@ rmw_create_publisher(
         DDS_DYNAMIC_DATA_MEMBER_ID_UNSPECIFIED, \
         value); \
     if (status != DDS_RETCODE_OK) { \
-      printf(#METHOD_NAME "() failed. Status = %d\n", status); \
+      printf(#METHOD_NAME "(%s) failed. Status = %d\n", member->name_, status); \
       throw std::runtime_error("set member failed"); \
     } \
   }
@@ -339,7 +343,7 @@ rmw_create_publisher(
       DDS_DYNAMIC_DATA_MEMBER_ID_UNSPECIFIED, \
       value SUFFIX); \
     if (status != DDS_RETCODE_OK) { \
-      printf(#METHOD_NAME "() failed. Status = %d\n", status); \
+      printf(#METHOD_NAME "(%s) failed. Status = %d\n", member->name_, status); \
       throw std::runtime_error("set member failed"); \
     } \
   }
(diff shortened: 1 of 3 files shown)


tfoote:
lgtm

rmw_connext_xtypes_dynamic_cpp/src/functions.cpp
```diff
@@ -907,18 +1172,98 @@ rmw_create_service(
...
+  DDSDataReader * request_datareader = replier->get_request_datareader();
+
+  ConnextDynamicServiceInfo * server_info = new ConnextDynamicServiceInfo();
+  server_info->replier_ = replier;
+  server_info->request_datareader_ = request_datareader;
+  server_info->response_type_support_ = response_type_support;
+  server_info->request_members_ = request_members;
+  server_info->response_members_ = response_members;
+
+  rmw_service_t * service = new rmw_service_t;
```
esteve: Please add a comment about using the allocator functions in the future.

esteve:
+1

rmw_connext_xtypes_dynamic_cpp/src/functions.cpp
dirk-thomas: Added in f8f5e8be2bf0d004a10ad729b1a4235f056465b0 - also for the static Connext implementation.

dirk-thomas:
Passing test: http://54.183.26.131:8080/view/ros2/job/ros2_batch_ci_linux/62/testReport/(root)/test_request_response__rmw_connext_xtypes_dynamic_cpp/

rmw_connext_xtypes_dynamic_cpp/src/functions.cpp
```diff
@@ -889,16 +1095,76 @@ rmw_take_response(
...
+    return RMW_RET_ERROR;
+  }
+
+  ConnextDynamicClientInfo * client_info = \
+    static_cast<ConnextDynamicClientInfo *>(client->data);
+  connext::Requester<DDS_DynamicData, DDS_DynamicData> * requester = client_info->requester_;
+
+  connext::LoanedSamples<DDS::DynamicData> replies = requester->take_replies(1);
+  typedef connext::LoanedSamples<DDS::DynamicData>::iterator iterator;
+  for (iterator it = replies.begin(); it != replies.end(); ++it) {
```
wjwwood: You could use the for range loop or even just `auto it = replies.begin()`. It's not important to change this one, but we are using C++11, we should make the most of it!

rmw_connext_xtypes_dynamic_cpp/src/functions.cpp
```diff
@@ -889,16 +1095,76 @@ rmw_take_response(
...
+    printf("  rmw_take_response() got a reply\n");
```
wjwwood: Also, should we leave this kind of thing in? (underscores the need for a proper logging system)

rmw_connext_xtypes_dynamic_cpp/src/functions.cpp
```diff
@@ -889,16 +1095,76 @@ rmw_take_response(
...
+  }
+  if (replies.begin() != replies.end() && replies.begin()->info().valid_data) {
+    _take(&replies.begin()->data(), ros_response, client_info->response_members_);
+
+    rmw_request_id_t & req_id = *(reinterpret_cast<rmw_request_id_t *>(ros_request_header));
+    req_id.sequence_number = ((int64_t)replies.begin()->related_identity().sequence_number.high)
+      << 32 | replies.begin()->related_identity().sequence_number.low;
```
wjwwood: The use of parenthesis here seems strange to me. I would have thought that you would have included the `<< 32` in the parenthesis. Was this intentional?

wjwwood:
Other than the comments, lgtm.

rmw_connext_xtypes_dynamic_cpp/src/functions.cpp
dirk-thomas: This is currently consistent with all the other rmw implementation. But it should for sure be removed at some point.

dirk-thomas: Nvm, this should actually be removed right now.

rmw_connext_xtypes_dynamic_cpp/src/functions.cpp
dirk-thomas: Just copied from the static implementation: https://github.com/ros2/rmw_connext/blob/1aece5657d69a34e8c808943105e3bb45a66c2ad/rosidl_typesupport_connext_cpp/resource/srv_ServiceTypeSupport.cpp.template#L126

I will wrap it in parenthesis in both locations.

rmw_connext_xtypes_dynamic_cpp/src/functions.cpp
dirk-thomas: Gone.

PR #28, opened 2015-04-30
rename rmw connext dynamic package

dirk-thomas:


dirk-thomas:
Since a test build shows no change at all (as it should be) I will just go ahead and merge these... :wink:

PR #33, opened 2015-05-19
use queue size to ensure history depth

dirk-thomas:
Connects to ros2/ros2#42

@esteve @tfoote @wjwwood Please review.

PR #34, opened 2015-05-19
workaround unbounded strings with dynamic data api

dirk-thomas:
Connects to ros2/ros2#42

Builds on top of #33.


tfoote:
+1

wjwwood:
+1

dirk-thomas:
To stay inter operable with the static Connext version the size should be 255 (a577f5e3d3c7d301746168b4e9d0a306ecabfa62).

PR #36, opened 2015-05-20
implement static array support

dirk-thomas:
Connects to ros2/ros2#42


tfoote:
+1, looks like a pain but makes sense

esteve:
+1

PR #38, opened 2015-05-22
enable checking message dependencies

dirk-thomas:
Connects to ros2/rosidl#12


rosidl_typesupport_connext_cpp/bin/rosidl_typesupport_connext_cpp
```diff
@@ -3,6 +3,8 @@
 import argparse
 import sys
 
+from rosidl_cmake import extract_message_types
+from rosidl_parser import UnknownMessageType
```
tfoote: This one needs rosidl_parser as a dependency. It will probably not fail due to transitive, but should be there.

dirk-thomas: Absolutely, added an explicit exec dependency. Same for `rosidl_typesupport_opensplice_cpp`.

tfoote:
+1

wjwwood:
+1

PR #40, opened 2015-06-02
fix warnings

wjwwood:
Fixes some warnings (hopefully) about unknown pragmas and arguments. See:

https://github.com/ros2/rmw_connext/commit/ee775b84fb304ed0d7c3413240566816ec1711c8#commitcomment-11295655

Connects to ros2/ros2#53

### DIFF
--- a/rosidl_typesupport_connext_cpp/cmake/rosidl_typesupport_connext_cpp_generate_interfaces.cmake
+++ b/rosidl_typesupport_connext_cpp/cmake/rosidl_typesupport_connext_cpp_generate_interfaces.cmake
@@ -68,13 +68,13 @@ endforeach()
 
 # If not on Windows, disable some warnings with Connext's generated code
 if(NOT WIN32)
-  if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
     set(_connext_compile_flags
       "-Wno-tautological-compare "
       "-Wno-return-type-c-linkage "
       "-Wno-deprecated-register "
     )
-  elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
+  elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
     set(_connext_compile_flags
       "-Wno-unused-variable "
       "-Wno-unused-but-set-variable "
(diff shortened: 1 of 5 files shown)


rosidl_typesupport_connext_cpp/cmake/rosidl_typesupport_connext_cpp_generate_interfaces.cmake
```diff
@@ -61,11 +61,21 @@ endforeach()
...
-    "-Wno-deprecated-register"
-  )
+  if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
+    set(_connext_compile_flags
+      "-Wno-deprecated-register "
+      "-Wno-mismatched-tags "
+      "-Wno-return-type-c-linkage "
+      "-Wno-sometimes-uninitialized "
+      "-Wno-tautological-compare "
+      "-Wno-unused-variable "
```
dirk-thomas: The string values should not have a trailing space inside. Same below for the GNU flags.

wjwwood: When I did this it would pass the flags to gcc all concatenated together. See (search for `unrecognized command line option`):

http://54.183.26.131:8080/job/ros2_batch_ci_linux/124/consoleFull
```
cc1plus: warning: unrecognized command line option "-Wno-tautological-compare-Wno-return-type-c-linkage-Wno-deprecated-register" [enabled by default]
```

dirk-thomas: The `set` command creates a list here. This implies that the items are separated by a semicolon. Therefore each item should not have a trailing space.

The additional problem here is that the replace call in line 79 has a bug and does not what it is supposed to do, namely replace these semicolons with space. That is because the value of the replace call is not quoted. It should be:

    string(REPLACE ";" " " _connext_compile_flags "${_connext_compile_flags}")

dirk-thomas:
This will also suppress warnings in the code the `rosidl_typesupport_connext_cpp` generates (which is not being generated by Connext's `idlpp`). Is there a better way to maintain the warnings in "our" code but suppress the warnings in Connext code?

wjwwood:
You can put `#pragma GCC diagnostic ignore "..."` around our code where we include their headers. We can also be more selective on which source files we add these flags to.

dirk-thomas:
Sounds good. I can update the CMake block after we merged the refactoring PR and only apply the properties for the generated files coming from the Connext code generator.

PR #43, opened 2015-06-22
fix using same type code multiple times, check for all error cases explicity

dirk-thomas:
This fixes ros2/system_tests#13 for both rmw connext implemenations. Additionally it check for all error cases explicitly and reports them via the return code and the rwm error string.

Please review and double check that I have not missed any cases which need to be checked.

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -75,7 +75,7 @@ rmw_init()
 {
   DDSDomainParticipantFactory * dpf_ = DDSDomainParticipantFactory::get_instance();
   if (!dpf_) {
-    rmw_set_error_string("  init() could not get participant factory");
+    rmw_set_error_string("failed to get participant factory");
     return RMW_RET_ERROR;
   }
   return RMW_RET_OK;
@@ -86,7 +86,7 @@ rmw_create_node(const char * name)
 {
   DDSDomainParticipantFactory * dpf_ = DDSDomainParticipantFactory::get_instance();
   if (!dpf_) {
-    rmw_set_error_string("  create_node() could not get participant factory");
+    rmw_set_error_string("failed to get participant factory");
     return NULL;
   }
 
@@ -94,7 +94,7 @@ rmw_create_node(const char * name)
   DDS_DomainParticipantQos participant_qos;
   DDS_ReturnCode_t status = dpf_->get_default_participant_qos(participant_qos);
   if (status != DDS_RETCODE_OK) {
-    rmw_set_error_string("  create_node() could not get participant qos");
+    rmw_set_error_string("failed to get default participant qos");
     return NULL;
   }
   status = DDSPropertyQosPolicyHelper::add_property(
@@ -103,7 +103,7 @@ rmw_create_node(const char * name)
     "0",
     DDS_BOOLEAN_FALSE);
   if (status != DDS_RETCODE_OK) {
-    rmw_set_error_string("  create_node() could not add qos property");
+    rmw_set_error_string("failed to add qos property");
     return NULL;
   }
 
@@ -114,7 +114,7 @@ rmw_create_node(const char * name)
     domain, participant_qos, NULL,
     DDS_STATUS_MASK_NONE);
   if (!participant) {
-    rmw_set_error_string("  create_node() could not create participant");
+    rmw_set_error_string("failed to create participant");
     return NULL;
   }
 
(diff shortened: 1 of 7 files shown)


tfoote:
These error strings don't automatically provide backtrace functionality.  Do we want to inject the implementation name into the error string to accelerate finding the source code for the error?

dirk-thomas:
Enhancing `rmw_set_error_string` to include e.g. filename, line number sounds like a help feature. I guess this would go into `rmw`.

wjwwood:
That's a good idea we can change `rmw_set_error_string` to be `rmw_set_error_state(char * file, int line, char * message)` and then add something like this to `rmw`:

```c
#define RMW_SET_ERROR_STRING(message) rmw_set_error_state(__FILE__, __LINE__, (message));
```

tfoote:
+1 for the current changes. I opened ros2/rmw#16 for the enhancement.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -132,61 +132,87 @@ rmw_create_publisher(
...
 
-  DDS_TopicQos default_topic_qos;
-  status = participant->get_default_topic_qos(default_topic_qos);
-  if (status != DDS_RETCODE_OK) {
-    rmw_set_error_string("failed to get default topic qos");
-    // printf("get_default_topic_qos() failed. Status = %d\n", status);
-    return NULL;
-  }
+  DDSTopic * topic;
+  DDSTopicDescription * topic_description = participant->lookup_topicdescription(topic_name);
```
wjwwood: What's this change about? Is this just avoiding creating the topic more than once?

rmw_connext_cpp/src/functions.cpp
```diff
@@ -132,61 +132,87 @@ rmw_create_publisher(
...
+      topic_name, type_name.c_str(), default_topic_qos, NULL,
+      DDS_STATUS_MASK_NONE
+      );
+    if (!topic) {
+      rmw_set_error_string("failed to create topic");
+      return NULL;
+    }
+  } else {
+    DDS_Duration_t timeout = DDS_Duration_t::from_seconds(0);
+    topic = participant->find_topic(topic_name, timeout);
```
wjwwood: Does this mean non-blocking or blocking?

rmw_connext_cpp/src/functions.cpp
dirk-thomas: This is necessary to fix ros2/system_tests#13. Otherwise the second time a topic is created results in a crash of `rmw_connext_*`.

rmw_connext_cpp/src/functions.cpp
dirk-thomas: The timeout of zero has no special semantic. It simply returns immediately.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -185,22 +192,42 @@ DDS_TypeCode * create_type_code(
...
+            // error string was set within the function
+            return NULL;
+          }
         }
         break;
       default:
-        printf("unknown type id %u\n", member->type_id_);
-        throw std::runtime_error("unknown type");
+        throw std::runtime_error(
+          std::string("unknown type id ") + std::to_string(member->type_id_));
```
wjwwood: Should this not be a `rmw_set_error_string` and `return NULL`?

dirk-thomas: Currently all `default` cases throw an exception. It is similar to other code where we do the same when an *enum* is not in the range we expect it to be. Not sure if we want to keep it that way.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -286,33 +340,42 @@ rmw_create_publisher(
...
-    topic_name, type_name.c_str(), default_topic_qos, NULL,
-    DDS_STATUS_MASK_NONE
-    );
-  if (!topic) {
-    rmw_set_error_string("  create_topic() could not create topic");
-    return NULL;
+    topic = participant->create_topic(
+      topic_name, type_name.c_str(), default_topic_qos, NULL,
+      DDS_STATUS_MASK_NONE
+      );
```
wjwwood: [style] As far as I understand our style guide, this should either have two less spaces in front of it or it should be on the previous line.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -995,39 +1163,88 @@ void _take(DDS_DynamicData * dynamic_data, void * ros_message,
...
+          bool success = _take(&sub_dynamic_data, sub_ros_message, sub_members);
+          status = dynamic_data->unbind_complex_member(sub_dynamic_data);
+          if (!success) {
+            rmw_set_error_string("failed to take sub message");
+            return false;
+          }
+          if (status != DDS_RETCODE_OK) {
+            rmw_set_error_string("failed to unbind complex member");
+            return false;
+          } \
```
wjwwood: Why does this end with a `\`?

rmw_connext_dynamic_cpp/src/functions.cpp
dirk-thomas: Good point. Fixed.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -995,39 +1163,88 @@ void _take(DDS_DynamicData * dynamic_data, void * ros_message,
...
         }
         break;
       default:
-        printf("unknown type id %u\n", member->type_id_);
-        throw std::runtime_error("unknown type");
+        throw std::runtime_error(
+          std::string("unknown type id ") + std::to_string(member->type_id_));
```
wjwwood: Should this be an exception? Is it caught before passing through the `rmw` interface?

rmw_connext_dynamic_cpp/src/functions.cpp
dirk-thomas: That was copy-n-pasted from a macro. Fixed.

rmw_connext_dynamic_cpp/src/functions.cpp
dirk-thomas: I replaced all exceptions with setting error strings / returning error codes.

wjwwood: Well, we could print to stderr and exit if you really don't want to miss a default switch case:

http://man7.org/linux/man-pages/man3/exit.3.html

dirk-thomas: This way it is at least consistent. And any code using the rmw API should check the return code anyway.

If we ever add another enum this way it will not break existing code if it written defensively.

wjwwood:
Pending my suggestion of using print to stderr and exit for default switch statements, it lgtm. This is a great improvement, thanks @dirk-thomas.

rmw_connext_dynamic_cpp/src/functions.cpp
wjwwood: Ok, sounds good to me.

esteve:
+1

dirk-thomas:
I will go ahead with the merge then.

Anybody volunteering to work on a similar PR for `rmw_opensplice`?

wjwwood:
If I end up working on https://github.com/ros2/rmw/issues/16 I'd update opensplice along with necessary changes to connext. Maybe after tomorrow's meeting I'll spend some time on that.

PR #44, opened 2015-06-23
add flag to ignore local publications

dirk-thomas:
Required for ros2/ros1_bridge#1

Please review.

### DIFF
--- a/rmw_connext_dynamic_cpp/src/functions.cpp
+++ b/rmw_connext_dynamic_cpp/src/functions.cpp
@@ -741,6 +741,7 @@ struct CustomSubscriberInfo
 {
   DDSDynamicDataTypeSupport * dynamic_data_type_support_;
   DDSDynamicDataReader * dynamic_reader_;
+  bool ignore_local_publications;
   DDS_TypeCode * type_code_;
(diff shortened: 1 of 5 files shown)


wjwwood:
+1

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -1263,15 +1266,38 @@ rmw_take(const rmw_subscription_t * subscription, void * ros_message, bool * tak
...
+      DDS_Octet * sender_element = &(sender_guid.value[i]);
+      DDS_Octet * receiver_element = &(((DDS_Octet *)&receiver_instance_handle)[i]);
+      if (*sender_element != *receiver_element) {
+        ignore_sample = false;
+        break;
+      }
+    }
+  }
+
+  if (!ignore_sample) {
```
dirk-thomas: If `ignore_sample` is `true` and `success` is `true` the code must set `taken` to `false`. Currently it is not a problem since the only code using the API initializes `taken` with `false` before calling but shouldn't need to.

esteve:
+1

tfoote:
+1

PR #45, opened 2015-06-29
Refactoring of create_* and destroy_* functions

wjwwood:
This has been done to resolve some teardown related errors and generally to improve the state of memory management in these implementations.

I've done several things here:

- Implemented destroy functions where they were missing before.
- Refactored create functions to properly unroll actions which had already been made in the event of an error later in the create function. This uses the "goto fail" pattern that is popular in the Linux kernel.
- Wrapped C++ constructor and destructor calls in try-catch cases to prevent escalation of exceptions through the C api.
- Replaced use of new and delete where ever possible with rmw_allocate to allow us more control over allocation in the future.
- Other minor improvements.

Things that are yet to be done (covered by TODO's):

- Some error propagation is not possible in the current setup (or error strings may get overwritten by other errors before returning).
- I think we might be leaking calls to `register_type` on the participant, but it's not clear if we must unregister them or if so, when we should unregister them.
- Servers and clients new more, msg specific, refactoring to replace use of new and (currently missing) delete calls to create "repliers" and "requesters".

Connects ros2/ros2#69

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -12,8 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <exception>
 #include <iostream>
+#include <sstream>
 #include <stdexcept>
+#include <string>
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wall"
@@ -29,9 +32,21 @@
 #include <rmw/error_handling.h>
 #include <rmw/types.h>
 
+#include <rmw/impl/cpp/macros.hpp>
+
 #include "rosidl_typesupport_connext_cpp/identifier.hpp"
 #include <rosidl_typesupport_connext_cpp/message_type_support.h>
 #include <rosidl_typesupport_connext_cpp/service_type_support.h>
+
+inline std::string
+_create_type_name(
+  const message_type_support_callbacks_t * callbacks,
+  const std::string & sep)
+{
+  return
+    std::string(callbacks->package_name) +
+    "::" + sep + "::dds_::" + callbacks->message_name + "_";
+}
 
 extern "C"
 {
@@ -40,12 +55,14 @@ const char * rti_connext_identifier = "connext_static";
 
 struct ConnextStaticPublisherInfo
 {
+  DDSPublisher * dds_publisher_;
   DDSDataWriter * topic_writer_;
   const message_type_support_callbacks_t * callbacks_;
 };
 
 struct ConnextStaticSubscriberInfo
 {
+  DDSSubscriber * dds_subscriber_;
   DDSDataReader * topic_reader_;
   bool ignore_local_publications;
   const message_type_support_callbacks_t * callbacks_;
@@ -120,7 +137,6 @@ rmw_create_node(const char * name)
   DDS_DomainId_t domain = 0;
 
   DDSDomainParticipant * participant = dpf_->create_participant(
-    //domain, DDS_PARTICIPANT_QOS_DEFAULT, NULL,
     domain, participant_qos, NULL,
     DDS_STATUS_MASK_NONE);
   if (!participant) {
@@ -128,11 +144,58 @@ rmw_create_node(const char * name)
     return NULL;
   }
 
-  rmw_node_t * node_handle = new rmw_node_t;
+  rmw_node_t * node_handle = rmw_node_allocate();
+  if (!node_handle) {
+    rmw_set_error_string("failed to allocate memory for node handle");
+    if (dpf_->delete_participant(participant) != DDS_RETCODE_OK) {
+      std::stringstream ss;
+      ss << "leaking participant while handling failue at " <<
+        __FILE__ << ":" << __LINE__;
+      (std::cerr << ss.str()).flush();
+    }
+    return NULL;
+  }
   node_handle->implementation_identifier = rti_connext_identifier;
   node_handle->data = participant;
 
   return node_handle;
+}
+
+rmw_ret_t
+rmw_destroy_node(rmw_node_t * node)
+{
+  if (!node) {
+    rmw_set_error_string("node handle is null");
+    return RMW_RET_ERROR;
+  }
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    node handle,
+    node->implementation_identifier, rti_connext_identifier,
+    return RMW_RET_ERROR)
+
+  DDSDomainParticipantFactory * dpf_ = DDSDomainParticipantFactory::get_instance();
+  if (!dpf_) {
+    rmw_set_error_string("failed to get participant factory");
+    return RMW_RET_ERROR;
+  }
+
+  DDSDomainParticipant * participant = static_cast<DDSDomainParticipant *>(node->data);
+  if (!participant) {
+    rmw_set_error_string("participant handle is null");
+  }
+  // This unregisters types and destroys topics which were shared between
+  // publishers and subscribers and could not be cleaned up in the delete functions.
+  participant->delete_contained_entities();
+
+  DDS_ReturnCode_t ret = dpf_->delete_participant(participant);
+  if (ret != DDS_RETCODE_OK) {
+    rmw_set_error_string("failed to delete participant");
+    return RMW_RET_ERROR;
+  }
+
+  rmw_node_free(node);
+
+  return RMW_RET_OK;
 }
 
 rmw_publisher_t *
@@ -146,20 +209,20 @@ rmw_create_publisher(
     rmw_set_error_string("node handle is null");
     return NULL;
   }
-  if (node->implementation_identifier != rti_connext_identifier) {
-    rmw_set_error_string("node handle is not from this rmw implementation");
-    return NULL;
-  }
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    node handle,
+    node->implementation_identifier, rti_connext_identifier,
+    return NULL)
+
   if (!type_support) {
     rmw_set_error_string("type support handle is null");
     return NULL;
   }
-  if (type_support->typesupport_identifier !=
-    rosidl_typesupport_connext_cpp::typesupport_connext_identifier)
-  {
-    rmw_set_error_string("type support is not from this rmw implementation");
-    return NULL;
-  }
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    type support handle,
+    type_support->typesupport_identifier,
+    rosidl_typesupport_connext_cpp::typesupport_connext_identifier,
+    return NULL)
 
   DDSDomainParticipant * participant = static_cast<DDSDomainParticipant *>(node->data);
   if (!participant) {
@@ -173,62 +236,73 @@ rmw_create_publisher(
     rmw_set_error_string("callbacks handle is null");
     return NULL;
   }
-  std::string type_name = std::string(callbacks->package_name) + "::msg::dds_::" +
-    callbacks->message_name + "_";
+  std::string type_name = _create_type_name(callbacks, "msg");
+  // Past this point, a failure results in unrolling code in the goto fail block.
+  rmw_publisher_t * publisher = nullptr;
+  bool registered;
+  DDS_PublisherQos publisher_qos;
+  DDS_ReturnCode_t status;
+  DDSPublisher * dds_publisher = nullptr;
+  DDSTopic * topic;
+  DDSTopicDescription * topic_description = nullptr;
+  DDS_DataWriterQos datawriter_qos;
+  DDSDataWriter * topic_writer = nullptr;
+  void * buf = nullptr;
+  ConnextStaticPublisherInfo * publisher_info = nullptr;
+  // Begin initializing elements
+  publisher = rmw_publisher_allocate();
+  if (!publisher) {
+    rmw_set_error_string("failed to allocate publisher");
+    goto fail;
+  }
 
-  bool registered = callbacks->register_type(participant, type_name.c_str());
+  registered = callbacks->register_type(participant, type_name.c_str());
   if (!registered) {
     rmw_set_error_string("failed to register type");
-    return NULL;
+    goto fail;
   }
 
-
-  DDS_PublisherQos publisher_qos;
-  DDS_ReturnCode_t status = participant->get_default_publisher_qos(publisher_qos);
+  status = participant->get_default_publisher_qos(publisher_qos);
   if (status != DDS_RETCODE_OK) {
     rmw_set_error_string("failed to get default publisher qos");
-    return NULL;
+    goto fail;
   }
 
-  DDSPublisher * dds_publisher = participant->create_publisher(
+  dds_publisher = participant->create_publisher(
     publisher_qos, NULL, DDS_STATUS_MASK_NONE);
   if (!dds_publisher) {
     rmw_set_error_string("failed to create publisher");
-    return NULL;
+    goto fail;
   }
 
-
-  DDSTopic * topic;
-  DDSTopicDescription * topic_description = participant->lookup_topicdescription(topic_name);
+  topic_description = participant->lookup_topicdescription(topic_name);
   if (!topic_description) {
     DDS_TopicQos default_topic_qos;
     status = participant->get_default_topic_qos(default_topic_qos);
     if (status != DDS_RETCODE_OK) {
       rmw_set_error_string("failed to get default topic qos");
-      return NULL;
+      goto fail;
     }
 
     topic = participant->create_topic(
       topic_name, type_name.c_str(), default_topic_qos, NULL, DDS_STATUS_MASK_NONE);
     if (!topic) {
       rmw_set_error_string("failed to create topic");
-      return NULL;
+      goto fail;
     }
   } else {
     DDS_Duration_t timeout = DDS_Duration_t::from_seconds(0);
     topic = participant->find_topic(topic_name, timeout);
     if (!topic) {
       rmw_set_error_string("failed to find topic");
-      return NULL;
+      goto fail;
     }
   }
 
-
-  DDS_DataWriterQos datawriter_qos;
   status = participant->get_default_datawriter_qos(datawriter_qos);
   if (status != DDS_RETCODE_OK) {
     rmw_set_error_string("failed to get default datawriter qos");
-    return NULL;
+    goto fail;
   }
 
   // ensure the history depth is at least the requested queue size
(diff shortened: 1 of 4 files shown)


wjwwood:
Working on a similar pr for opensplice now.

esteve:
+1

tfoote:
Overall it looks good. I think all the `goto fail` calls could be replaced with lambda functions for cleanlyness.

wjwwood:
Maybe the lambda would be cleaner, but I'm not convinced of that since you'd have to capture all of the variables, not sure if a catch all capture would work. Either way, this is like my 16th iteration on this change set and I'm not particularly motivated to make such an intrusive non-functional change :smile:.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -320,111 +454,283 @@ rmw_create_publisher(
...
+    ddts, buf,
+    goto fail,
+    DDSDynamicDataTypeSupport, type_code, DDS_DYNAMIC_DATA_TYPE_PROPERTY_DEFAULT);
+  buf = nullptr;  // Only free the casted pointer; don't need the buf pointer anymore.
   status = ddts->register_type(participant, type_name.c_str());
   if (status != DDS_RETCODE_OK) {
     rmw_set_error_string("failed to register type");
-    return NULL;
+    // Delete ddts to prevent the goto fail block from trying to unregister_type.
+    TRY_DESTRUCTOR_FROM_WITHIN_FAILURE(ddts->, DDSDynamicDataTypeSupport)
```
jacquelinekay: Seeing the dangling dereference arrow as input to the `TRY_DESTRUCTOR` macros is jarring for me. Are trying to leave the option open for another dereference operator besides `->`?

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -320,111 +454,283 @@ rmw_create_publisher(
...
+    }
+    custom_publisher_info->type_code_ = nullptr;
+    rmw_free(custom_publisher_info);
+  }
+  publisher->data = nullptr;
+  rmw_publisher_free(publisher);
+  return RMW_RET_OK;
 }
 
 #define SET_PRIMITIVE_VALUE(TYPE, METHOD_NAME) \
```
jacquelinekay: macro at the top of the file?

wjwwood: Maybe? but I didn't put that one there.

rmw_connext_dynamic_cpp/src/functions.cpp
wjwwood: > Answered my own question, looks like you use operator . at later points in the code

^, I didn't want to make two, really long named macros, but we could do that too. Also I could just put the whole destructor statement in there, but this is where I ended up.

wjwwood: Ok, I moved the macros to a common header in `rmw` and I changed them such that they pass the entire destructor statement (no dangling `->` or `.`).

rmw_connext_cpp/src/functions.cpp
```diff
@@ -119,13 +134,46 @@ rmw_create_node(const char * name)
...
   node_handle->implementation_identifier = rti_connext_identifier;
   node_handle->data = participant;
 
   return node_handle;
 }
 
+rmw_ret_t
+rmw_destroy_node(rmw_node_t * node)
+{
+  if (node->implementation_identifier != rti_connext_identifier) {
```
dirk-thomas: This must check if node is not null before.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -119,13 +134,46 @@ rmw_create_node(const char * name)
     return NULL;
   }
 
-  rmw_node_t * node_handle = new rmw_node_t;
+  rmw_node_t * node_handle = rmw_node_allocate();
+  if (!node_handle) {
+    rmw_set_error_string("failed to allocate memory for node handle");
+    if (dpf_->delete_participant(participant) != DDS_RETCODE_OK) {
+      std::cerr << "leaking participant while handling failue at " <<
+        __FILE__ << ":" << __LINE__;
```
dirk-thomas: We should avoid streaming to stdout / stderr since this will potentially interleave output when being using in multiple threads.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -164,62 +212,73 @@ rmw_create_publisher(
     rmw_set_error_string("callbacks handle is null");
     return NULL;
   }
-  std::string type_name = std::string(callbacks->package_name) + "::msg::dds_::" +
-    callbacks->message_name + "_";
+  std::string type_name = _create_type_name(callbacks, "msg");
+  // Past this point, a failure results in unrolling code in the goto fail block.
```
dirk-thomas: Instead of using `goto` here we could use a `while` loop and break from it.

rmw_connext_dynamic_cpp/src/functions.cpp
dirk-thomas: I would keep the macros as close to where they are being used as possible.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -102,33 +117,79 @@ rmw_create_node(const char * name)
...
+  rmw_node_t * node = rmw_node_allocate();
   node->implementation_identifier = rti_connext_dynamic_identifier;
   node->data = participant;
   return node;
 }
 
+rmw_ret_t
+rmw_destroy_node(rmw_node_t * node)
+{
+  if (node->implementation_identifier != rti_connext_dynamic_identifier) {
```
dirk-thomas: Check `node` before.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -1692,51 +2284,208 @@ rmw_create_client(
...
+rmw_ret_t
+rmw_destroy_client(rmw_client_t * client)
+{
+  if (!client) {
+    rmw_set_error_string("client handle is null");
+    return RMW_RET_ERROR;
+  }
+
+  auto result = RMW_RET_OK;
+  ConnextDynamicClientInfo * client_info = static_cast<ConnextDynamicClientInfo *>(client->data);
```
dirk-thomas: Check rmw implementation type of client handle before.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -2096,24 +2964,49 @@ rmw_destroy_service(rmw_service_t * service)
...
-}
-
-rmw_ret_t
-rmw_destroy_client(rmw_client_t * client)
-{
-  if (!client) {
-    rmw_set_error_string("client handle is null");
-    return RMW_RET_ERROR;
+  auto result = RMW_RET_OK;
+  auto service_info = static_cast<ConnextDynamicServiceInfo *>(service->data);
```
dirk-thomas: Check rmw implementation type of service handle before.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -847,10 +1152,14 @@ rmw_destroy_client(rmw_client_t * client)
     return RMW_RET_ERROR;
   }
 
+  auto result = RMW_RET_OK;
   // TODO(esteve): de-allocate Requester and response DataReader
-  delete static_cast<ConnextStaticClientInfo *>(client->data);
-  delete client;
-  return RMW_RET_OK;
+  TRY_DESTRUCTOR(
+    static_cast<ConnextStaticClientInfo *>(client->data)->~ConnextStaticClientInfo(),
```
dirk-thomas: Check rmw implementation type of client handle before.

rmw_connext_cpp/src/functions.cpp
wjwwood: I looked around for the "best" practice for this pattern and I saw this recommendation as an alternative to `goto fail`:

```
do {
  a = ...;
  if (!a) break;
  b = ...;
  if (!b) break;
  // and so on
  return success;
} while (false);
// clean up
return failure;
```

Which seems like what you're suggesting, but it seemed to be that people generally recommended the goto fail style. The example everyone kept pointing back to was the Linux kernel.

I'm open to changing it, but I'd like to know what others think. @codebot specifically, you've done more of this style of coding, what do you think?

jacquelinekay: `goto fail` is a security vulnerability if the programmer accidentally duplicates the `goto fail` statement:

https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/

Whereas the do/while loop alternative creates danger if the programmer forgets a single break statement, or duplicates a break statement.

I see problems with both approaches. +0, want to hear what Morgan thinks.

wjwwood: IMO, the "goto fail" bug that you linked to is a "failure to use `{}` with an `if` statement. As you said, it could easily be duplicated with `break`'s in a `while` loop.

rmw_connext_cpp/src/functions.cpp
wjwwood: This is done, but further up so this comment doesn't close.

wjwwood:
I've addressed all of the comments, pending the decision about `goto fail`.

rmw_connext_cpp/src/functions.cpp
wjwwood: Bump @codebot, if you cold weigh in here I can finish this pr up. Thanks.

dirk-thomas: A good discussion which mentions multiple pros / cons / alternatives: http://stackoverflow.com/questions/1073397/while1-break-instead-of-goto

wjwwood: Based on that link I'd say `goto fail` is more obvious, doesn't have problems when inside nesting loops, and can be expanded to have multiple recovery stages with different labels. Where as the `do {} while(false)` method can have problems when the interior code needs to break out of a loop but not for an error and it cannot have multiple recovery labels but instead requires conditionals to be set in the interior and evaluated in the recovery code.

So I'd say there's not sufficient reason to change away from the `goto fail`. Obligatory:

https://xkcd.com/292/

tfoote: Careful with your references. The raptors might get back into ROS if morgan's following this thread...

+1 for sticking with goto, the target name might be better a little bit more verbosely named though.

wjwwood: > the target name might be better a little bit more verbosely named though.

I assume you mean the label name? What would you suggest, `goto fail` seems pretty straight forward to me.

tfoote: Yes it is straight forward, but very hard to quickly introspect. Using the same target label means that you can't search/jump quickly without checking that you're in the same scope. There's 12 instances in this PR. 

xyz_cleanup where xyz are unique would seem to be a better pattern so that you can quickly do a search and audit that that specific set of goto's without having to mentally filter the other scopes.

wjwwood: Ok, I can look at prefixing the function name to the label (lots of busy work), but honestly when ever you encounter a `goto fail;` the very next instance of `fail:` should be the label it goes to and so searching for it should be straight forward.

rmw_connext_dynamic_cpp/src/functions.cpp
esteve: I prefer macros to be at the top of files, but that's just a personal preference.

rmw_connext_cpp/src/functions.cpp
esteve: I'm ok with just `goto fail`, it's a widely used pattern in the Linux kernel.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -29,10 +33,21 @@
...
 #include <rosidl_typesupport_connext_cpp/message_type_support.h>
 #include <rosidl_typesupport_connext_cpp/service_type_support.h>
 
+inline std::string
+_create_type_name(
+  const message_type_support_callbacks_t * callbacks,
+  const std::string & sep)
+{
+  return std::string(callbacks->package_name) +
+         "::" + sep + "::dds_::" + callbacks->message_name + "_";
```
dirk-thomas: Wrong indent.

rmw_connext_cpp/src/functions.cpp
dirk-thomas: I would vote for sticking to `fail`.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -35,12 +39,24 @@
...
 #include "rosidl_typesupport_introspection_cpp/visibility_control.h"
 
+ROSIDL_TYPESUPPORT_INTROSPECTION_CPP_LOCAL
+inline std::string
+_create_type_name(
+  const rosidl_typesupport_introspection_cpp::MessageMembers * members,
+  const std::string & sep)
+{
+  return std::string(members->package_name_) + "::" + sep + "::dds_::" + members->message_name_ +
+         "_";
```
dirk-thomas: Wrong indent.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -111,33 +127,91 @@ rmw_create_node(const char * name)
...
     return NULL;
   }
-
+  // Past this point, a failure results in unrolling code in the goto fail block.
   DDS_StructMemberSeq struct_members;
   DDS_ExceptionCode_t ex = DDS_NO_EXCEPTION_CODE;
-  DDS_TypeCode * type_code = factory->create_struct_tc(type_name.c_str(), struct_members, ex);
+  DDS_TypeCode * type_code = nullptr;
+  // Start initializing elements.
+  type_code = factory->create_struct_tc(type_name.c_str(), struct_members, ex);
```
dirk-thomas: Why separate these two lines?

dirk-thomas:
All tests related to `staticarrayprimitives` and `dynamicarrayprimitives` fail. Looking at one of them (publisher_subscriber_cpp__staticarrayprimitives__rmw_connext_dynamic_cpp) it fails with the following error:

    DDS_DynamicDataTypeSupport_initialize:typecode error 5 on content_type
    raising: could not create subscription: failed to register type

Therefore I looked at the type creation function - especially for arrays. For the subscriber: https://github.com/ros2/rmw_connext/blob/962fe936019c0fa081cb3c0d08661b85edb3502b/rmw_connext_dynamic_cpp/src/functions.cpp#L302-L360

The code first create the type support for the base type and then if it is an array creates the type support for the array / sequence which is based on the base type. But the PR removes the base type support which is required for the array / sequence type support: https://github.com/ros2/rmw_connext/commit/5e7236f3eb4b749ebb1c4bf0e3af2bd7c745be89#diff-d8fc519447c043a7165e1ff89d1c2088R325

I have fixed the type support in commit 6cf72f33f898dffe5c2614ebc72a0153170cd54d.

wjwwood:
Ok thanks, I guess this will be captured in the CI job you just started?:

http://ci.ros2.org/job/ros2_batch_ci_linux/43/

rmw_connext_cpp/src/functions.cpp
wjwwood: I'll stick with `goto fail` then.

rmw_connext_cpp/src/functions.cpp
wjwwood: I ran uncrustify over this already and I wouldn't call that indent wrong, but it is an alternative indent that we don't usually use. In this case I think it looks fine, in other cases like this:

```c++
  some_really_long_named_function_that_does_stuff(arg1,
                                                  arg2,
                                                  arg3);
```

I think that style doesn't make sense, but there has to be some room for developer discretion in these types of things.

rmw_connext_dynamic_cpp/src/functions.cpp
wjwwood: In spite of what I've said in the other one, I updated both instances in cc82120, though I think both the change I made and the alternative:

```c++
  return std::string(members->package_name_) + "::" + sep + "::dds_::" + members->message_name_ +
    "_";
```

don't look as good.

rmw_connext_dynamic_cpp/src/functions.cpp
wjwwood: It's obviously not necessary, but it just followed the pattern I had going in other functions where I initialized everything that could fail to null and then set them and then checked them in the failure block.

rmw_connext_dynamic_cpp/src/functions.cpp
wjwwood: I don't really mind one way or the other, but I'm not going to move them in my pr.

rmw_connext_dynamic_cpp/src/functions.cpp
dirk-thomas: I was wondering since a few lines above (https://github.com/ros2/rmw_connext/pull/45/files#diff-d8fc519447c043a7165e1ff89d1c2088R196) the patch does the exact opposite change.

wjwwood: Basically I was just operating on the principal that if a failure to set the value results in a goto fail then it should be first initialized to null first, but that could be more sophisticatedly done as "if a failure to set the value results in a goto fail then it should be initialized to null first, unless it is the first thing to be set", but I didn't do that. Does it really matter? Why are we wasting time on a non-functional change? I wish we could move away from detailed back and forth on miniature issues like this. I changed it in: be37f96

jacquelinekay:
Jenkins looks great... +1

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -111,33 +127,90 @@ rmw_create_node(const char * name)
...
+    return RMW_RET_ERROR;
+  }
+
+  DDS_ExceptionCode_t ex;
+  factory->delete_tc(type_code, ex);
+  if (ex != DDS_NO_EXCEPTION_CODE) {
+    rmw_set_error_string("failed to delete type code struct");
+    return RMW_RET_ERROR;
+  }
+  // TODO(wjwwood): is this sufficient clean up of the type code?
```
dirk-thomas: Is `destroy_type_code` invoked for every type code which is being created? If yes, I would assume that should be sufficient.

Or are there currently cases which it not invoked yet?

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -1052,14 +1634,27 @@ rmw_create_subscription(const rmw_node_t * node,
...
           i + 1); \
         if (status != DDS_RETCODE_OK) { \
-          delete[] values; \
+          if (values) { \
+            /* First call the destructor for each element, then free the whole block. */ \
+            for (size_t i = 0; i < array_size; ++i) { \
+              RMW_TRY_DESTRUCTOR(values[i].~DDS_TYPE(), DDS_TYPE, ) \
+            } \
+            rmw_free(values); \
+          } \
```
dirk-thomas: Since this block is repeated numerous time should it be moved into a separate macro?

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -2017,6 +2855,7 @@ rmw_take_response(
     return RMW_RET_ERROR;
   }
 
+  // TODO(wjwwood): return this loan, maybe it's not necessary? I don't see a function to do it.
```
dirk-thomas: The documentation (http://community.rti.com/rti-doc/510/ndds/doc/html/api_cpp/classconnext_1_1Requester.html#aeccadc80d171a7ae6573369e9d9bfbbb) indicates that the loan needs to be returned explicitly using `return_loan` or implicitly by the destructor.

rmw_connext_dynamic_cpp/src/functions.cpp
wjwwood: I suppose it could, I'll leave that for a later improvement.

rmw_connext_dynamic_cpp/src/functions.cpp
wjwwood: I think I just forgot to remove this after a refactoring (hidden due to the squash). I removed it in de87fd8

rmw_connext_dynamic_cpp/src/functions.cpp
wjwwood: Ok, so the destructor should take care of it, removing the TODO's: 91a2715

PR #47, opened 2015-07-09
fix handling of array of non-primitive types

dirk-thomas:
This is necessary to correctly handle arrays of non-primitive types which are covered by the new tests in ros2/system_tests#15.

This depends on PR ros2/rosidl_dds#23 which extends the introspection interface.

### DIFF
--- a/rmw_connext_dynamic_cpp/src/functions.cpp
+++ b/rmw_connext_dynamic_cpp/src/functions.cpp
@@ -196,8 +196,16 @@ DDS_TypeCode * create_type_code(
         break;
       case rosidl_typesupport_introspection_cpp::ROS_TYPE_MESSAGE:
         {
+          if (!member->members_) {
+            rmw_set_error_string("members handle is null");
+            return NULL;
+          }
           const::rosidl_typesupport_introspection_cpp::MessageMembers * sub_members =
             (const::rosidl_typesupport_introspection_cpp::MessageMembers *)member->members_->data;
+          if (!sub_members) {
+            rmw_set_error_string("sub members handle is null");
+            return NULL;
+          }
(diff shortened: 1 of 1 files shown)


rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -587,6 +599,50 @@ rmw_create_publisher(
...
+    rmw_set_error_string("failed to bind complex member"); \
+    return false; \
+  } \
+  void * sub_ros_message = (void *)((char *)ros_message + member->offset_); \
+  if (!member->members_) { \
+    rmw_set_error_string("members handle is null"); \
+    return false; \
+  } \
+  const::rosidl_typesupport_introspection_cpp::MessageMembers * sub_members = \
+    (const::rosidl_typesupport_introspection_cpp::MessageMembers *)member->members_->data; \
```
jacquelinekay: should this be a `static_cast<const::rosidl_typesupport_introspection_cpp::MessageMembers *>` instead?

wjwwood:
+1. I'll have to rebase my change set, but that's fine.

rmw_connext_dynamic_cpp/src/functions.cpp
dirk-thomas: That would be better. Since the code block was only moved in the PR from below into a macro I didn't change it. I will address the C casts in another round.

jacquelinekay:
+1

tfoote:
+1

rmw_connext_dynamic_cpp/src/functions.cpp
dirk-thomas: I created a follow up PR for the const correctness: #48

PR #48, opened 2015-07-10
replace regular casts with C++ style casts

dirk-thomas:
Beside const correctness the code should also use reinterpret casts instead of C-style casts.

@wjwwood We can hold this PR until #45 has been merged if that is preferred?

### DIFF
--- a/rmw_connext_dynamic_cpp/src/functions.cpp
+++ b/rmw_connext_dynamic_cpp/src/functions.cpp
@@ -275,8 +275,9 @@ DDS_TypeCode * create_type_code(
             RMW_SET_ERROR_MSG("members handle is null");
             return NULL;
           }
-          const::rosidl_typesupport_introspection_cpp::MessageMembers * sub_members =
-            (const::rosidl_typesupport_introspection_cpp::MessageMembers *)member->members_->data;
+          auto sub_members =
+            static_cast<const::rosidl_typesupport_introspection_cpp::MessageMembers *>(
+            member->members_->data);
           if (!sub_members) {
             RMW_SET_ERROR_MSG("sub members handle is null");
             return NULL;
@@ -469,14 +470,14 @@ rmw_create_publisher(
     rosidl_typesupport_introspection_cpp::typesupport_introspection_identifier,
     return NULL)
 
-  DDSDomainParticipant * participant = (DDSDomainParticipant *)node->data;
+  auto participant = static_cast<DDSDomainParticipant *>(node->data);
   if (!participant) {
     RMW_SET_ERROR_MSG("participant handle is null");
     return NULL;
   }
 
-  const rosidl_typesupport_introspection_cpp::MessageMembers * members =
-    (const rosidl_typesupport_introspection_cpp::MessageMembers *)type_support->data;
+  auto members = static_cast<const rosidl_typesupport_introspection_cpp::MessageMembers *>(
+    type_support->data);
   if (!members) {
     RMW_SET_ERROR_MSG("members handle is null");
     return NULL;
@@ -739,13 +740,13 @@ rmw_destroy_publisher(rmw_node_t * node, rmw_publisher_t * publisher)
     publisher->implementation_identifier, rti_connext_dynamic_identifier,
     return RMW_RET_ERROR)
 
-  DDSDomainParticipant * participant = (DDSDomainParticipant *)node->data;
+  auto participant = static_cast<DDSDomainParticipant *>(node->data);
   if (!participant) {
     RMW_SET_ERROR_MSG("participant handle is null");
     return RMW_RET_ERROR;
   }
 
-  CustomPublisherInfo * custom_publisher_info = (CustomPublisherInfo *)publisher->data;
+  auto custom_publisher_info = static_cast<CustomPublisherInfo *>(publisher->data);
   if (custom_publisher_info) {
     DDSDynamicDataTypeSupport * ddts = custom_publisher_info->dynamic_data_type_support_;
     if (ddts) {
@@ -806,11 +807,11 @@ rmw_destroy_publisher(rmw_node_t * node, rmw_publisher_t * publisher)
 }
 
 #define SET_PRIMITIVE_VALUE(TYPE, METHOD_NAME) \
-  const TYPE value = *((const TYPE *)((const char *)ros_message + member->offset_)); \
+  const TYPE * value = reinterpret_cast<const TYPE *>(static_cast<const char *>(ros_message) + member->offset_); \
   DDS_ReturnCode_t status = dynamic_data->METHOD_NAME( \
     NULL, \
     i + 1, \
-    value); \
+    *value); \
   if (status != DDS_RETCODE_OK) { \
     RMW_SET_ERROR_MSG("failed to set primitive value using " #METHOD_NAME); \
     return false; \
@@ -820,11 +821,11 @@ rmw_destroy_publisher(rmw_node_t * node, rmw_publisher_t * publisher)
   const TYPE * ros_values = nullptr; \
   size_t array_size; \
   if (member->array_size_ && !member->is_upper_bound_) { \
-    ros_values = (const TYPE *)((const char *)ros_message + member->offset_); \
+    ros_values = reinterpret_cast<const TYPE *>(static_cast<const char *>(ros_message) + member->offset_); \
     array_size = member->array_size_; \
   } else { \
-    auto untyped_vector = (const void *)((const char *)ros_message + member->offset_); \
-    auto vector = reinterpret_cast<const std::vector<TYPE> *>(untyped_vector); \
+    const void * untyped_vector = static_cast<const char *>(ros_message) + member->offset_; \
+    auto vector = static_cast<const std::vector<TYPE> *>(untyped_vector); \
     ros_values = vector->data(); \
     array_size = vector->size(); \
   }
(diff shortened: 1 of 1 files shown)


wjwwood:
Yeah, I plan to finalize that one today. I still have a couple failing test, but I think I know what needs to change.

wjwwood:
I say we go ahead and merge this one.

wjwwood:
nvm lets hold.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -327,7 +327,7 @@ rmw_create_publisher(
   }
 
   const rosidl_typesupport_introspection_cpp::MessageMembers * members =
-    (rosidl_typesupport_introspection_cpp::MessageMembers *)type_support->data;
+    (const rosidl_typesupport_introspection_cpp::MessageMembers *)type_support->data;
```
esteve: I'd rather use C++ casts instead.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -449,7 +449,7 @@ rmw_create_publisher(
 }
 
 #define SET_PRIMITIVE_VALUE(TYPE, METHOD_NAME) \
-  TYPE value = *((TYPE *)((char *)ros_message + member->offset_)); \
+  const TYPE value = *((const TYPE *)((const char *)ros_message + member->offset_)); \
```
esteve: Same here (re casts).

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -460,14 +460,14 @@ rmw_create_publisher(
   }
 
 #define ARRAY_SIZE_AND_VALUES(TYPE) \
-  TYPE * ros_values = nullptr; \
+  const TYPE * ros_values = nullptr; \
   size_t array_size; \
   if (member->array_size_ && !member->is_upper_bound_) { \
-    ros_values = (TYPE *)((char *)ros_message + member->offset_); \
+    ros_values = (const TYPE *)((const char *)ros_message + member->offset_); \
```
esteve: Same here (re casts).

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -460,14 +460,14 @@ rmw_create_publisher(
...
     array_size = member->array_size_; \
   } else { \
-    auto untyped_vector = (void *)((char *)ros_message + member->offset_); \
-    auto vector = reinterpret_cast<std::vector<TYPE> *>(untyped_vector); \
+    auto untyped_vector = (const void *)((const char *)ros_message + member->offset_); \
```
esteve: Casts.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -525,14 +525,14 @@ rmw_create_publisher(
       size_t array_size; \
       if (member->array_size_ && !member->is_upper_bound_) { \
         array_size = member->array_size_; \
-        auto ros_values = (TYPE *)((char *)ros_message + member->offset_); \
+        auto ros_values = (const TYPE *)((const char *)ros_message + member->offset_); \
```
esteve: Casts.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -525,14 +525,14 @@ rmw_create_publisher(
...
         values = new DDS_TYPE[array_size]; \
         for (size_t j = 0; j < array_size; ++j) { \
           values[j] = ros_values[j]; \
         } \
       } else { \
-        auto untyped_vector = (void *)((char *)ros_message + member->offset_); \
-        auto vector = reinterpret_cast<std::vector<TYPE> *>(untyped_vector); \
+        auto untyped_vector = (const void *)((const char *)ros_message + member->offset_); \
```
esteve: Casts.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -855,7 +855,7 @@ rmw_create_subscription(const rmw_node_t * node,
   }
 
   const rosidl_typesupport_introspection_cpp::MessageMembers * members =
-    (rosidl_typesupport_introspection_cpp::MessageMembers *)type_support->data;
+    (const rosidl_typesupport_introspection_cpp::MessageMembers *)type_support->data;
```
esteve: Casts.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -1786,7 +1789,7 @@ rmw_create_client(
   }
 
   const rosidl_typesupport_introspection_cpp::ServiceMembers * service_members =
-    (rosidl_typesupport_introspection_cpp::ServiceMembers *)type_support->data;
+    (const rosidl_typesupport_introspection_cpp::ServiceMembers *)type_support->data;
```
esteve: Casts.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -2135,7 +2138,7 @@ rmw_create_service(
...
```
esteve: Casts.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -718,9 +718,9 @@ bool _publish(DDS_DynamicData * dynamic_data, const void * ros_message,
             }
             size_t array_size = member->size_function(untyped_member);
             for (size_t j = 0; j < array_size; ++j) {
-              const void * const_sub_ros_message = member->get_function(untyped_member, j);
+              const void * const_sub_ros_message = member->get_const_function(untyped_member, j);
               // offset message pointer since the macro adds the member offset to it
-              void * ros_message = (void *)((char *)const_sub_ros_message - member->offset_);
+              const void * ros_message = (const void *)((const char *)const_sub_ros_message - member->offset_);
```
esteve: Casts.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -697,13 +697,13 @@ bool _publish(DDS_DynamicData * dynamic_data, const void * ros_message,
       case rosidl_typesupport_introspection_cpp::ROS_TYPE_MESSAGE:
         {
           if (member->is_array_) {
-            auto untyped_member = (void *)((char *)ros_message + member->offset_);
+            auto untyped_member = (const void *)((const char *)ros_message + member->offset_);
```
esteve: Casts.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -587,7 +587,7 @@ rmw_create_publisher(
         return false; \
       } \
     } else { \
-      TYPE value = *((TYPE *)((char *)ros_message + member->offset_)); \
+      TYPE value = *((const TYPE *)((const char *)ros_message + member->offset_)); \
```
esteve: Casts.

esteve:
Besides the comments about casts, this looks good to me.

dirk-thomas:
@esteve This PR is only about const correctness. Once William PRs are merged this can be merged and then all C style cast should also be replaced.

esteve:
@dirk-thomas if that's the case, please revert this to keep the original cast:

https://github.com/ros2/rmw_connext/blob/const_correctness/rmw_connext_dynamic_cpp/src/functions.cpp#L612

wjwwood:
I rebased this.

Based on everything I've read so far, if you're casting to/from a `void *`, then `static_cast` is the most correct thing to do, but C++ style casting is not explicitly needed for casting from a `void *`, C style would suffice. And assigning into a `void *` requires no casting at all. So I'd propose updating this pull request to do the following for conversion to/from `void *`:

- Convert all C style casts to C++ style `static_cast` (preserving constness where needed).
- Re-evaluate the use of `reinterpret_cast` in these cases and consider converting them to `static_cast`.
- Consider dropping casts all together when assigning into a `void *`.

I want to have a consensus on the approach before proceeding with the required changes.

dirk-thomas:
Fully agreed:

* never use C-style casts
* favor `static_cast` over `reinterpret_cast` wherever possible
* dropping casts when not necessary

I can update my PR if you want to work on other stuff.

wjwwood:
Sounds good.

wjwwood:
Looks like this has been rebased but not refactored w.r.t. c++ cast style. I moved it back to in progress.

dirk-thomas:
I replaced all remaining regular casts with C++ style casts in a second commit.

rmw_connext_dynamic_cpp/src/functions.cpp
jacquelinekay: yeah, shouldn't this be `const rosidl_typesupport_introspection_cpp::ServiceMembers * service_members = reinterpret_cast<const rosidl_typesupport_introspection_cpp::ServiceMembers *>(type_support->data);`?

dirk-thomas: Good point. That's another advantage for the C++ style casts: you can actually search for them reliable. I found a few more like this and converted them in 21887a32ba027efb5e52dea6ec346a110d4e6312.

jacquelinekay:
+1

wjwwood:
+1

PR #50, opened 2015-07-16
reset pointers to null after destroying them

dirk-thomas:
Otherwise most of the checks for `0` don't make sense.

This is required for the other rmw implementations as well.


wjwwood:
This is a nice clean up, but why is this necessary?

wjwwood:
+1

dirk-thomas:
If code has a `node` handle which gets destroyed it can check that the `data` pointer is already a `nullptr` and stop using it / throw with a reasonable message instead of segfaulting.

PR #51, opened 2015-07-16
make typesupport namespace unique for each rmw implementation

dirk-thomas:
Otherwise message lib A from type support X might use symbols from message B from type support Y.


wjwwood:
+1

It really sucks that we have to keep side stepping behavior of the linker because of how we over link the message package implementations.

PR #52, opened 2015-07-17
remove unnecessary try_placement / try_destructor calls on primitive values in arrays

dirk-thomas:
Related to #45 introducing those calls.

### DIFF
--- a/rmw_connext_dynamic_cpp/src/functions.cpp
+++ b/rmw_connext_dynamic_cpp/src/functions.cpp
@@ -798,24 +798,12 @@ rmw_destroy_publisher(rmw_node_t * node, rmw_publisher_t * publisher)
       ARRAY_SIZE_AND_VALUES(TYPE) \
       DDS_TYPE * values = nullptr; \
       if (array_size > 0) { \
-        /* Allocate the memory, but do not do placement new yet. */ \
         values = reinterpret_cast<DDS_TYPE *>(rmw_allocate(sizeof(DDS_TYPE) * array_size)); \
         if (!values) { \
           rmw_set_error_string("failed to allocate memory"); \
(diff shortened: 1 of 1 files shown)


rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -824,13 +812,7 @@ rmw_destroy_publisher(rmw_node_t * node, rmw_publisher_t * publisher)
...
         array_size, \
         values); \
-      if (values) { \
-        /* First call the destructor for each element, then free the whole block. */ \
-        for (size_t i = 0; i < array_size; ++i) { \
-          RMW_TRY_DESTRUCTOR(values[i].~DDS_TYPE(), DDS_TYPE, ) \
-        } \
-        rmw_free(values); \
-      } \
+      rmw_free(values); \
```
wjwwood: I guess there is an assumption here that `rmw_free` behaves like `free` when it is given a nullptr. We'll just have to remember to do the same if we provide an alternative allocation strategy.

wjwwood:
+1, I guess I just assumed since they were in macros and used new it was better to just try the constructor/destructor, but looking at the use of the macros now, it's clear that isn't the case.

If you're feeling defensive, you could do a static assert to prevent the macros from being misused on a non-trivial type in the future with something like this in the macros:

```c++
static_assert(std::is_trivial<DDS_TYPE>, "cannot be used with non-trivial type");
```

wjwwood:
(I fixed a typo in the example snippet above)

dirk-thomas:
I will for now refrain from adding checks to the macros. They are already very complicated. Anybody shuffling around code in this area needs a very good understanding of the side effects anyway.

PR #53, opened 2015-07-18
unify qos properties across all data readers and writers

dirk-thomas:
It fixes the missing qos properties for the requesters / repliers.
It adds new qos properties to handle unbounded data dynamically.

CI job: http://ci.ros2.org/job/ros2_batch_ci_linux/56/


wjwwood:
+1

PR #57, opened 2015-07-21
change take flags to be in line with rmw opensplice implementation

dirk-thomas:
Change flag for take to `DDS_ANY_SAMPLE_STATE`.


jacquelinekay:
Just curious, are you testing the demo with connext and if so, did you see the same message ordering problem before these changes?

dirk-thomas:
I test the demo with all nine combinations of camera node / bridge. The ordering problem only happened when using OpenSplice. For more details about other issues please see ros2/ros2#56

PR #60, opened 2015-07-24
Pass timeout to rmw_wait

jacquelinekay:
Connects to ros2/ros2#73

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -927,7 +927,7 @@ rmw_wait(rmw_subscriptions_t * subscriptions,
   rmw_guard_conditions_t * guard_conditions,
   rmw_services_t * services,
   rmw_clients_t * clients,
-  bool non_blocking)
+  rmw_time_t * wait_timeout)
 {
   DDSWaitSet waitset;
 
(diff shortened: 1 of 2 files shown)


rmw_connext_cpp/src/functions.cpp
```diff
@@ -883,7 +883,7 @@ rmw_wait(rmw_subscriptions_t * subscriptions,
   rmw_guard_conditions_t * guard_conditions,
   rmw_services_t * services,
   rmw_clients_t * clients,
-  bool non_blocking)
+  rmw_time_t * wait_timeout)
```
dirk-thomas: The function must check that the pointer is valid (if it stay a pointers).
(Actually it should also check all the other pointers coming in.)

Same below and for OpenSplice.

dirk-thomas:
This needs a rebase since the commit list contains other commits.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -994,8 +994,17 @@ rmw_wait(rmw_subscriptions_t * subscriptions,
...
+  if (!wait_timeout) {
+    timeout = DDS_DURATION_INFINITE;
+  } else {
+    non_blocking = wait_timeout->sec == 0 && wait_timeout->nsec == 0;
+    timeout.sec = wait_timeout->sec;
+    timeout.nanosec = wait_timeout->nsec;
+  }
   DDS_ReturnCode_t status = DDS_RETCODE_TIMEOUT;
+
   while (DDS_RETCODE_TIMEOUT == status) {
```
jacquelinekay: Double posting comment from https://github.com/ros2/rmw_opensplice/pull/55:

As I was writing a test for this, I realized that the existing logic of this loop is not what I thought it was.

Old behavior:

* Get "blocking" or "non-blocking in rclcpp.
* In rmw implementation:

* if non-blocking, call rmw_wait with a timeout of zero, and don't loop if timed out.
* if blocking: while return code is timeout, wait for 1 second. throw if error, continue if retcode was not timeout. The executor will always block indefinitely inside of rmw_wait.

Here's what I think the new behavior should be:

Get timeout from rclcpp.
* If NULL: block indefinitely in rmw_wait, by passing `DURATION_INFINITE` to `WaitSet::Wait`.
* If positive: pass the timeout value to `WaitSet::wait` and return if timed out.
* If non-blocking: same behavior as before.

dirk-thomas:
+1

esteve:
+1

wjwwood:
+1

tfoote:
+1 with rebase/squash

PR #62, opened 2015-07-29
fix returning from rmw_wait while any samples have not been taken yet

dirk-thomas:
The patch currently only fixes the subscriptions. The same fix is needed for services as well.

This makes the new tests from ros2/system_tests#20 pass: http://ci.ros2.org/job/ros2_batch_ci_linux/106/


wjwwood:
+1

PR #65, opened 2015-08-03
Topic information

dirk-thomas:
Connects to ros2/ros1_bridge#3

The `rmw_connext_cpp` implementation is ready for review.

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -16,9 +16,12 @@
 #include <exception>
 #include <iostream>
 #include <limits>
+#include <map>
+#include <set>
 #include <sstream>
 #include <stdexcept>
 #include <string>
+#include <vector>
 
 #ifndef _WIN32
 # pragma GCC diagnostic push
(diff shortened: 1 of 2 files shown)


rmw_connext_cpp/src/functions.cpp
```diff
@@ -1630,4 +1846,246 @@ rmw_send_response(
...
+  }
+  if (!node_info->subscriber_listener) {
+    RMW_SET_ERROR_MSG("subscriber listener handle is null");
+    return RMW_RET_ERROR;
+  }
+
+  // combine publisher and subscriber information
+  std::map<std::string, std::set<std::string>> topics_with_multiple_types;
+  for (auto it : node_info->publisher_listener->topic_names_and_types) {
+    for (auto jt : it.second) {
```
wjwwood: Maybe use `auto & jt` here to avoid copying the string.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -1630,4 +1846,246 @@ rmw_send_response(
...
+
+  // reformat type name
+  std::string substr = "::msg::dds_::";
+  for (auto it : topics) {
+    size_t substr_pos = it.second.find(substr);
+    if (it.second[it.second.size() - 1] == '_' && substr_pos != std::string::npos)
+    {
+      it.second = it.second.substr(0, substr_pos) + "/" +
+        it.second.substr(substr_pos + substr.size(), it.second.size() - substr_pos - substr.size() - 1);
+      topics[it.first] = it.second;
```
wjwwood: I don't think this should be necessary if you get a reference to the pair from the `topics` map.

wjwwood:
lgtm

rmw_connext_cpp/src/functions.cpp
dirk-thomas: Good point. Done in 04cef82c99e31c35cd58d0786d422ccca15ccb98

rmw_connext_cpp/src/functions.cpp
dirk-thomas: Also done in 04cef82c99e31c35cd58d0786d422ccca15ccb98

tfoote:
+1

esteve:
+1

esteve:
BTW, here's an idea, we could expose this information as a parameter too. I haven't given much thought to it, though.

PR #66, opened 2015-08-03
Added read_condition for services and clients

esteve:
This should fix ros2/rclcpp#52 for services

Connects to ros2/rclcpp#52

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -88,6 +88,7 @@ struct ConnextStaticClientInfo
 {
   void * requester_;
   DDSDataReader * response_datareader_;
+  DDSReadCondition * read_condition_;
   const service_type_support_callbacks_t * callbacks_;
 };
 
@@ -95,6 +96,7 @@ struct ConnextStaticServiceInfo
 {
   void * replier_;
   DDSDataReader * request_datareader_;
+  DDSReadCondition * read_condition_;
   const service_type_support_callbacks_t * callbacks_;
 };
 
@@ -622,6 +624,7 @@ rmw_destroy_subscription(rmw_node_t * node, rmw_subscription_t * subscription)
     return RMW_RET_ERROR;
   }
   // TODO(wjwwood): need to figure out when to unregister types with the participant.
+  auto result = RMW_RET_OK;
   ConnextStaticSubscriberInfo * subscriber_info =
     (ConnextStaticSubscriberInfo *)subscription->data;
   if (subscriber_info) {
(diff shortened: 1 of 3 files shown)


esteve:
@dirk-thomas I applied the same fix as for subscriptions, but can't seem to make the client fire when the service sends back the response. I ran `rtiddsspy` and messages are sent between the service and the client, though. Would appreciate any feedback from you. Thanks.

esteve:
@dirk-thomas I think I found where the problem is, the client's read condition was not being attached to the waitset.

esteve:
Linux buildfarm:

http://ci.ros2.org/job/ros2_batch_ci_linux/137/

rmw_connext_cpp/src/functions.cpp
```diff
@@ -1242,6 +1221,13 @@ rmw_create_client(
     goto fail;
   }
 
+  read_condition = response_datareader->create_readcondition(
```
dirk-thomas: This needs to be cleaned up in the `fail` section.

Same below.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -1466,14 +1489,33 @@ rmw_destroy_service(rmw_service_t * service)
...
+  ConnextStaticServiceInfo * service_info = static_cast<ConnextStaticServiceInfo *>(service->data);
+
+  if (service_info) {
+    auto request_datareader = service_info->request_datareader_;
+    if (request_datareader) {
+      auto read_condition = service_info->read_condition_;
+      if (read_condition) {
+        if (request_datareader->delete_readcondition(read_condition) != DDS_RETCODE_OK) {
+          RMW_SET_ERROR_MSG("failed to delete readcondition");
+          return RMW_RET_ERROR;
```
dirk-thomas: This should only set the return code but continue to perform the other clean up operations.

Same below and above.

esteve: Fixed.

rmw_connext_cpp/src/functions.cpp
esteve: Fixed.

dirk-thomas:
+1

Has it been tested that this fixes ros2/system_tests#14

esteve:
Buildfarm job:

http://ci.ros2.org/job/ros2_batch_ci_linux/143/

dirk-thomas:
This CI job does cover the new parameter test yet.

wjwwood:
Code changes look good, waiting on the test.

tfoote:
+1, unfortunately this needs to be rebased now

esteve:
Rebased.

jacquelinekay:
I had to rebase this on master to get it to build on my machine (currently test ros2/system_tests#57)

jacquelinekay:
+1

PR #69, opened 2015-08-05
fix checks when deleting entities

dirk-thomas:


dirk-thomas:
I will go ahead and merge this trivial bug fix. I can take care of any feedback afterwards.

PR #72, opened 2015-08-12
Check error codes in attach_condition

jacquelinekay:
Improve the error message for RMW_SET_ERROR_MSG based on different DDS return codes for `Waitset::attach_condition`.

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -174,6 +174,21 @@ struct ConnextStaticServiceInfo
   const service_type_support_callbacks_t * callbacks_;
 };
 
+rmw_ret_t check_attach_condition_error(DDS::ReturnCode_t retcode)
+{
+  if (retcode == DDS_RETCODE_OK) {
+    return RMW_RET_OK;
+  }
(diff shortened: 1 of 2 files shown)


rmw_connext_cpp/src/functions.cpp
```diff
@@ -174,6 +174,23 @@ struct ConnextStaticServiceInfo
   const service_type_support_callbacks_t * callbacks_;
 };
 
+rmw_ret_t rmw_check_attach_condition_error(DDS_ReturnCode_t retcode)
```
dirk-thomas: Since this is not a function specified in the rmw interface please use a different name (e.g. without the `rmw_` prefix).

Same below and for OpenSplice.

tfoote:
+1

esteve:
+1

dirk-thomas:
+1

dirk-thomas:
This seems to break the build: http://ci.ros2.org/view/nightly/job/ros2_batch_ci_osx_nightly/9/console Please fix.

jacquelinekay:
ros2/rmw_connext/pull/85

PR #74, opened 2015-08-13
Fix cmake module for MSVC 2015

### LINKED ISSUE #73: CMake module is not working for MSVC 2015
The [FindConnext.cmake]() was not detecting my home made installation of RTI Connext which mimic the 5.1 and 5.2 RTI official directory layout. I would propose the following patch: 

```patch
diff --git a/connext_cmake_module/cmake/Modules/FindConnext.cmake b/connext_cmake_module/cmake/Modules/FindConnext.cmake
index 6180438..7fb3df5 100644
--- a/connext_cmake_module/cmake/Modules/FindConnext.cmake
+++ b/connext_cmake_module/cmake/Modules/FindConnext.cmake
@@ -166,27 +166,6 @@ if(NOT "${_NDDSHOME} " STREQUAL " ")
     endif()
   endwhile()
 
-  if(_matched_VS2015)
-    set(_i 0)
-    while(TRUE)
-      list(LENGTH _libs _length)
-      if(NOT ${_i} LESS ${_length})
-        break()
-      endif()
-      list(GET _libs ${_i} _lib)
-      set(_match TRUE)
-      string(FIND "${_lib}" "VS2015" _found)
-      if(NOT ${_found} EQUAL -1)
-        set(_match FALSE)
-      endif()
-      if(_match)
-        math(EXPR _i "${_i} + 1")
-      else()
-        list(REMOVE_AT _libs ${_i})
-      endif()
-    endwhile()
-  endif()
-
```

The patch removes an extra check for MSVC2015 which seems not needed to me since the check is already done in [previous code](https://github.com/ros2/rmw_connext/blob/master/connext_cmake_module/cmake/Modules/FindConnext.cmake#L151). And it is particularly buggy in [checking the presence VS2015 string](https://github.com/ros2/rmw_connext/blob/master/connext_cmake_module/cmake/Modules/FindConnext.cmake#L179). Removing the patch worked fine for me in my tests.

j-rivero:
Closes #73.


dirk-thomas:
Maybe @wjwwood can comment on why it was introduced in #8?

Assuming the block is not used any more all remaining instances of the variable `_matched_VS2015` should also be removed. Another question is since we won't support VS2013 anymore if we should also remove the detection of it here?

j-rivero:
> Assuming the block is not used any more all remaining instances of the variable _matched_VS2015 should also be removed.

Done.

> Another question is since we won't support VS2013 anymore if we should also remove the detection of it here?

Ok for me, but that will kill our Windows jenkins support until we install the 2015 version of the RTI Connext libraries. I can prepare a different pull request and we will merged it as soon as ros2/ros2#51 is solved.

j-rivero:
My [testing run with RTI community version](http://ci.ros2.org/job/ros2_batch_ci_windows/240/consoleFull#console-section-39) is failing probably because we have not merged this PR. @wjwwood, @dirk-thomas green light to merge?

dirk-thomas:
Can you please commit these changes to a branch in this repo and then pass the branch name to the CI job? That way we can test it and merge it after it is confirmed to work.

j-rivero:
> Can you please commit these changes to a branch in this repo and then pass the branch name to the CI job? That way we can test it and merge it after it is confirmed to work.

Done pull request #92.

PR #77, opened 2015-08-14
Get debug library versions in the cmake module

j-rivero:
After merging pull request #74 (sorry about it, pretty small changes).

The RTI Community edition 5.2 is building libraries using a "d" postfix when building in Debug mode. `CMAKE_BUILD_TYPE` is empty when running in ament, so I used the `CMAKE_BUILD_TYPE_INIT`, although it seems not officially documented. Let me know if there is any better alternative.


wjwwood:
I think `CMAKE_BUILD_TYPE` is always empty (None) unless explicitly set. Is it not sufficient to just check if `CMAKE_BUILD_TYPE` is `Debug`, and if it's not set then that's ok too?

j-rivero:
> I think CMAKE_BUILD_TYPE is always empty (None) unless explicitly set (regardless of whether or not ament is involved).

I think so too. I was more thinking if ament was setting the CMAKE_BUILD_TYPE when releasing packages, but don't think so.

> Is it not sufficient to just check if CMAKE_BUILD_TYPE is Debug, and if it's not set then that's ok too?

Rethinking a bit more on this, I would say that the main difference when using the Visual Studio Generator is that it will generate vcxproj files which support both types (Release/Debug) and you call them when run msbuild. This is different than in *NIX, when we are building just one makefile. So this PR is probably wrong and we need to find the way of implementing the logic to get the proper libraries inside the vcxproj.

wjwwood:
Ok, well do what you think is right. I was just curious as to why special logic was needed here.

j-rivero:
Closing. For more information please follow #86

dirk-thomas:
Closing in favor of #89.

PR #81, opened 2015-08-15
check if rtiddsgen_server is runnable

dirk-thomas:
If the Java program is not runnable (e.g. no Java runtime environment available) it does not set the variable (and therefore depending code falls back using the non-server generator.

Failed in http://ci.ros2.org/job/ros2_batch_ci_osx/129/
Worked in http://ci.ros2.org/job/ros2_batch_ci_osx/130/


wjwwood:
+1

PR #83, opened 2015-08-17
Accept QoS profiles via an XML file

esteve:
Connect to ros2/ros2#86


dirk-thomas:
Closing due to https://github.com/ros2/ros2/issues/86#issuecomment-140549073

PR #84, opened 2015-08-17
update topic information when entities go away

dirk-thomas:


wjwwood:
> No description provided.

Can you explain what you've changed and why it's now possible? AFAIR, you didn't see anyway to accomplish this. What changed? I see you're comparing instance handles now, what was the issue before?

dirk-thomas:
I got feedback from RTI on my question. Before I tried to use the `key` but that one was not the same when receiving invalid data. Another approach uses a not documented function but that didn't work for me either. Now it uses the instance handle instead.

Instance handles are not hashable which requires going to all and compare with each one. That makes the patch not very efficient for many topics.

wjwwood:
+1

jacquelinekay:
+1

As a side note, there is a large and growing amount of duplicated code between `rmw_connext_cpp` and `rmw_connext_dynamic_cpp`. Is a shared library that provides classes such as `CustomDataReaderListener` worthwhile?

jacquelinekay:
+1

tfoote:
+1

dirk-thomas:
@jacquelinekay Sure, common code could be moved to a third package. Some of it might even be sharable between different vendors. There is also a lot of macro stuff in there which could be refactored...

Issue #86, opened 2015-08-19
Setup windows debug libraries in Connext_LIBRARIES (cmake module)


j-rivero:
The Visual Studio build of RTI Connext generates release and debug libraries (same using a "d" postfix). The issues is not critical as far as you don't need to debug the RTI Connext libraries. And the workaround would be to modify the `_excpected_library_base_names` in `FindConnext.cmake` package adding a `d` at the end.

I would say that the proper way of getting this defined in the Find cmake module would be to use:

```cmake
set(Connext_LIBRARIES optimized nddsc debug nddscd
                      optimized ...   debug ...)
```
Something similar is implemented in the [FindwxWindows.cmake module](https://github.com/Kitware/CMake/blob/master/Modules/FindwxWindows.cmake#L403)

When I tried a quick test for this layout ament is generating the error:

```
ament_export_libraries() package 'rmw_connext_cpp' passes the build configuration keyword 'debug' as the last exported library
```

Could be that some parsing/code in the ament package does not support this way of defining the libraries? I need to investigate about it.

dirk-thomas:
ament supports build configuration keywords. If you can point me to the modified version and what exactly you called I could take a look.

j-rivero:
> If you can point me to the modified version and what exactly you called I could take a look.

I run a quick test just adding this line at the end of the FindConnext.cmake module (just a stub for triggering the error).

```patch
@@ -249,6 +228,8 @@ if(Connext_FOUND AND NOT WIN32)
   list(APPEND Connext_LIBRARIES "pthread" "dl")
 endif()

+set(Connext_LIBRARIES optimized nddsc debug nddscd)
+
 include(FindPackageHandleStandardArgs)
 # Connext_HOME, Connext_LIBRARY_DIRS, and Connext_LIBRARY_DIR are not always se
 find_package_handle_standard_args(Connext
```

Install the connext_cmake_module and run `ament build --only rmw_connext_cpp`

```
 - Found rosidl_typesupport_connext_cpp: 0.0.0 (C:/Users/jrivero/code/ros2_build
/install/share/rosidl_typesupport_connext_cpp/cmake)
CMake Error at C:/Users/jrivero/code/ros2_build/install/share/ament_cmake_export
_libraries/cmake/ament_export_libraries.cmake:48 (message):
  ament_export_libraries() package 'rmw_connext_cpp' passes the build
  configuration keyword 'optimized' as the last exported library
Call Stack (most recent call first):
[...]

PR #89, opened 2015-08-19
use Connext debug libraries when building with CMAKE_BUILD_TYPE Debug

dirk-thomas:
Replacing #77 (following style). This is orthogonal to the use of the Community edition. It already works with the currently used professional version of Connext.

Since the CI jobs don't build with `CMAKE_BUILD_TYPE=Debug` there is no need to trigger jobs for this.

### DIFF
--- a/connext_cmake_module/cmake/Modules/FindConnext.cmake
+++ b/connext_cmake_module/cmake/Modules/FindConnext.cmake
@@ -66,11 +66,15 @@ function(_find_connext_ensure_libraries var expected_library_names library_paths
   set(${var} TRUE PARENT_SCOPE)
 endfunction()
 
+set(_lib_suffix "")
+if("${CMAKE_BUILD_TYPE} " STREQUAL "Debug ")
+  set(_lib_suffix "d")
+endif()
 set(_expected_library_base_names
-  "nddsc"
-  "nddscore"
-  "nddscpp"
-  "rticonnextmsgcpp"
+  "nddsc${_lib_suffix}"
+  "nddscore${_lib_suffix}"
+  "nddscpp${_lib_suffix}"
+  "rticonnextmsgcpp${_lib_suffix}"
 )
 
 set(_expected_library_names "")


dirk-thomas:
Is anybody using our custom Debian package of Connext 5.2? If yes, @esteve can you please release an updated Debian package which includes the debug libraries.

tfoote:
+1

I'm using our debian packages.

esteve:
I'm using the Debian packages too.

@j-rivero do you have time for updating the packages? I can show you how.

j-rivero:
@dirk-thomas: this PR implements my first approach. After thinking about what the Visual Studio cmake generator does and knowing that *after* running cmake the call to msbuild can invoke (using /p:Configuration) Release/Debug I would say that we need both logic into the vcxproj generated by cmake. And I think that the way to do it is ros2/rmw_connext#86

> @j-rivero do you have time for updating the packages? I can show you how.

Not today, a bit busy with gazebo stuff and Windows. But feel free to mail me the instructions and I will give a look tomorrow.

dirk-thomas:
I agree that ros2/rmw_connext#86 is the more flexible solution. I am not sure if we need it but if you like to look into it that would be good.

Still this simple approach should be merged as soon as we have a Debian package which contains debug libraries.

dirk-thomas:
@esteve Please catch up on this.

Moving back to in-progress for now.

dirk-thomas:
We need an ament specific flag to make `shared` the default (without requiring user interaction).

wjwwood:
From the meeting:

 - @esteve will ping people about the state of this.

esteve:
I've created a PR to add instructions for creating Debian packages for RTI Connext:

https://github.com/rticommunity/rticonnextdds-src/pull/13

let me know if there's anything that needs to be improved.

wjwwood:
@j-rivero talk with @esteve about testing the new instructions.

gerkey:
@dirk-thomas says that this is blocked by a lack of availability of debug Connext binaries (we've been unable to successfully build working versions of those ourselves).

dirk-thomas:
I will close this in favor of a more flexible solution which exports both libraries like this: `debug nddsd.so optimized ndds.so`

I have created a separate ticket for creating a Debian package which also includes debug libraries: osrf/rticonnextdds-src#4

PR #92, opened 2015-08-20
Fix cmake module for MSVC 2015

### LINKED ISSUE #73: CMake module is not working for MSVC 2015

j-rivero:
Closes #73. Same content that #74


esteve:
+1

j-rivero:
* [Failing build](http://ci.ros2.org/job/ros2_batch_ci_windows/240/) with the default branch
* [Passing build](http://ci.ros2.org/job/ros2_batch_ci_windows/247/console#console-section-6) with this branch

tfoote:
+1

wjwwood:
Why is this "in progress", is there more work to be done or should it be merged?

j-rivero:
> Why is this "in progress", is there more work to be done or should it be merged?

I don't think so.

wjwwood:
Sorry, I don't know which you mean. You don't think there is more work to be done or you don't think it should be merged?

We usually put things back into "in progress" if something comes up during the review process which indicates that it shouldn't be merged.

wjwwood:
I guess you mean it should be merged, since it was never in the "review" column.

wjwwood:
Sorry for the noise.

PR #94, opened 2015-08-21
disable deleting of DDSDynamicDataTypeSupport since it hasn't been unregistered before

dirk-thomas:


rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -996,22 +996,22 @@ rmw_create_publisher(
       }
     }
     if (ddts) {
-      // Cannot unregister type here, in case another topic is using the same type.
+      // TODO Cannot unregister and free type here, in case another topic is using the same type.
```
jacquelinekay: Just to check I'm understanding this: we currently don't have a way of checking if other topics are using the same type, so memory allocated for `ddts` will hang around until the node destructor is called, even if it's not being used.

dirk-thomas: Yes, we currently have no way to decide when to destroy a `Topic`. Therefore we can't unregister the `DDSDynamicDataTypeSupport` which is bound to the topic and therefore we can't destroy the `DDSDynamicDataTypeSupport`. That memory is not even recovered when the node is being destroyed but lost permanently.

jacquelinekay:
+1

wjwwood:
+1

tfoote:
+1

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -1900,25 +1900,25 @@ rmw_destroy_subscription(rmw_node_t * node, rmw_subscription_t * subscription)
...
-       */
       if (custom_subscription_info->dynamic_data) {
         if (ddts->delete_data(custom_subscription_info->dynamic_data) != DDS_RETCODE_OK) {
           RMW_SET_ERROR_MSG("failed to delete dynamic data");
           return RMW_RET_ERROR;
         }
+        custom_subscription_info->dynamic_data = nullptr;
+      }
+      std::string type_name = _create_type_name(custom_subscription_info->members_, "msg");
+      // TODO Cannot unregister and free type here, in case another topic is using the same type.
```
esteve: Do we know who wrote the original `TODO` comment? If so, could you add their username (i.e. `TODO(username))?

esteve:
+1

rmw_connext_dynamic_cpp/src/functions.cpp
dirk-thomas: I think it was William but I would rather not bother because now multiple people have updated the block already. Important is that we find this again - therefore I added the `TODO`.

jacquelinekay: It was Dirk, he actually added the TODO.

I noticed uncrustify used to give a style error if there was a TODO without a username, but no longer does.

PR #98, opened 2015-08-25
add support for gid in rmw

wjwwood:
Connects to ros2/ros2#91.

See: https://github.com/ros2/rmw/pull/35

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -189,11 +189,17 @@ struct ConnextStaticNodeInfo
   CustomSubscriberListener * subscriber_listener;
 };
 
+typedef struct ConnextStaticPublisherGID
+{
+  DDS_InstanceHandle_t publication_handle;
+} ConnextStaticPublisherGID;
+
 struct ConnextStaticPublisherInfo
 {
   DDSPublisher * dds_publisher_;
   DDSDataWriter * topic_writer_;
   const message_type_support_callbacks_t * callbacks_;
+  rmw_gid_t publisher_gid;
 };
 
 struct ConnextStaticSubscriberInfo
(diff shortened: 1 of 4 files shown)


rmw_connext_cpp/src/functions.cpp
```diff
@@ -1124,8 +1134,10 @@ rmw_destroy_subscription(rmw_node_t * node, rmw_subscription_t * subscription)
   return RMW_RET_OK;
 }
 
+RMW_LOCAL
```
dirk-thomas: Is that really necessary in the cpp file?

wjwwood: Prevents this symbol being visible in the symbol table of the `.so`. It expands to `__attribute__ ((visibility("hidden")))`. I suppose I could make it `static inline` instead.

jacquelinekay:
+1

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -2412,6 +2435,39 @@ rmw_take(const rmw_subscription_t * subscription, void * ros_message, bool * tak
...
+    return RMW_RET_ERROR;
+  }
+  DDS_InstanceHandle_t sending_publication_handle;
+  auto ret = _take_impl(subscription, ros_message, taken, &sending_publication_handle);
+  if (ret != RMW_RET_OK) {
+    // Error string is already set.
+    return RMW_RET_ERROR;
+  }
+
+  rmw_gid_t * sender_gid = const_cast<rmw_gid_t *>(&message_info->publisher_gid);
```
dirk-thomas: Neither `message_info` nor `publisher_gid` seems to be const. Why is the `const_cast` necessary here?

rosidl_typesupport_connext_cpp/resource/msg__type_support.cpp.template
```diff
@@ -297,6 +298,10 @@ take__@(spec.base_type.type)(
       }
     }
   }
+  if (sample_info.valid_data && sending_publication_handle != nullptr) {
```
dirk-thomas: `!= nullptr` not necessary.

dirk-thomas:
LGTM

tfoote:
+1

rmw_connext_dynamic_cpp/src/functions.cpp
wjwwood: They used to be, removed in 0488637

rosidl_typesupport_connext_cpp/resource/msg__type_support.cpp.template
wjwwood: Removed in 0488637

rmw_connext_cpp/src/functions.cpp
wjwwood: I just go rid of it, since the default visibility should be hidden anyways (this is done by adding a compiler flag in CMake). 0488637

wjwwood:
Squashed.

PR #99, opened 2015-08-27
declare unwanted but implicit dependency

dirk-thomas:
Until this implicitly dependency is removed (ros2/rmw_opensplice#39) it must be declared correctly declared in order to not break isolated builds.


jacquelinekay:
+1

tfoote:
+1

Issue #96, opened 2015-08-25
Client and service do not exchange messages


esteve:
See ros2/rclcpp#95

* `add_two_ints_server__rmw_connext_cpp` / `add_two_ints_client__rmw_connext_cpp` do not exchange any messages

esteve:
Testing this again, it seems this issue does not happen consistently.

jacquelinekay:
CI for the set of branches connected to this issue looks fine:

http://ci.ros2.org/job/ros2_batch_ci_linux/344/
http://ci.ros2.org/job/ros2_batch_ci_osx/263/
http://ci.ros2.org/job/ros2_batch_ci_windows/398/

PR #100, opened 2015-09-09
Fix default QoS settings for services

jacquelinekay:
Connects to ros2/rmw_connext#96

See ros2/rmw_opensplice/pull/82 for description.

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -229,56 +229,9 @@ rmw_create_publisher(
     }
   }
 
-  if (!get_datawriter_qos(participant, datawriter_qos)) {
+  if (!get_datawriter_qos(participant, qos_profile, datawriter_qos)) {
     // error string was set within the function
     goto fail;
-  }
-
-  switch (qos_profile.history) {
-    case RMW_QOS_POLICY_KEEP_LAST_HISTORY:
-      datawriter_qos.history.kind = DDS_KEEP_LAST_HISTORY_QOS;
-      break;
-    case RMW_QOS_POLICY_KEEP_ALL_HISTORY:
-      datawriter_qos.history.kind = DDS_KEEP_ALL_HISTORY_QOS;
-      break;
-    case RMW_QOS_POLICY_HISTORY_SYSTEM_DEFAULT:
-      break;
-    default:
-      RMW_SET_ERROR_MSG("Unknown QoS history policy");
-      goto fail;
-  }
-
-  switch (qos_profile.reliability) {
-    case RMW_QOS_POLICY_BEST_EFFORT:
-      datawriter_qos.reliability.kind = DDS_BEST_EFFORT_RELIABILITY_QOS;
-      break;
-    case RMW_QOS_POLICY_RELIABLE:
-      datawriter_qos.reliability.kind = DDS_RELIABLE_RELIABILITY_QOS;
-      break;
-    case RMW_QOS_POLICY_RELIABILITY_SYSTEM_DEFAULT:
-      break;
-    default:
-      RMW_SET_ERROR_MSG("Unknown QoS reliability policy");
-      goto fail;
-  }
-
-  if (qos_profile.depth != RMW_QOS_POLICY_DEPTH_SYSTEM_DEFAULT) {
-    datawriter_qos.history.depth = static_cast<DDS_Long>(qos_profile.depth);
-  }
-
-  // ensure the history depth is at least the requested queue size
-  assert(datawriter_qos.history.depth >= 0);
-  if (
-    datawriter_qos.history.kind == DDS::KEEP_LAST_HISTORY_QOS &&
-    static_cast<size_t>(datawriter_qos.history.depth) < qos_profile.depth
-  )
-  {
-    if (qos_profile.depth > (std::numeric_limits<DDS_Long>::max)()) {
-      RMW_SET_ERROR_MSG(
-        "failed to set history depth since the requested queue size exceeds the DDS type");
-      goto fail;
-    }
-    datawriter_qos.history.depth = static_cast<DDS_Long>(qos_profile.depth);
   }
 
   topic_writer = dds_publisher->create_datawriter(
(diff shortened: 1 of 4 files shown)


dirk-thomas:
I am not sure about the changes since the pub / sub equivalent of the code does something different (especially after #88):
* https://github.com/ros2/rmw_connext/commit/d9538a31ab8a5746956ed49e225ba8961595e6d0#diff-becdf01440f1757df19f4775b534e066R976
* https://github.com/ros2/rmw_connext/commit/d9538a31ab8a5746956ed49e225ba8961595e6d0#diff-becdf01440f1757df19f4775b534e066R991

Just from reading these two code parts it looks to me (on a first look) that they are contradicting each other.

I know this is not directly related to this PR but since this PR adds similar stuff for the services as for pub/sub I think those should be similar (or even the same).

jacquelinekay:
I'm not sure what you mean when you say these sections of code contradict each other.

These are the three differences I see:

1. The QoS configuration code for publishers sets the policy for a DataWriter object. The equivalent for subscriptions sets the policy for a DataReader object. The equivalent for services and clients sets both a DataReader and a DataWriter, because both clients and services have a DataReader and DataWriter object.
2. The service/client code casts the history policy depth as a DDS_Long, while the pub/sub code does not do any casting. I can take this out for the sake of similarity as I don't think it's required.
3. The service/client code sets the durability policy to TRANSIENT_LOCAL. I can also take this out for the sake of similarity, though I need to confirm first that it's not needed to fix the bug.

dirk-thomas:
The first code block **sets** the history depth of the writer to the requested value unconditionally (if `RMW_QOS_POLICY_DEPTH_SYSTEM_DEFAULT` was not requested).

The second block **increases** the history of the reader to the requested queue size.

The two logics are significantly different in behavior and I think they should be considered to to the same conceptional thing instead. Currently the resulting behavior is not obvious and different for readers and writers.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -934,6 +934,47 @@ rmw_create_client(
     goto fail;
   }
 
+  // For now, use the default QoS profile for services.
```
esteve: I'd move this code to `create_requester`, pass a `rmw_qos_profile_t` to `rmw_create_client` and default it to `rmw_qos_profile_services_default`.

jacquelinekay:
Oh, I see, Dirk is talking about lines 269-282 (create_publisher) and 599-612 (create_subscription).

https://github.com/ros2/rmw_connext/blob/master/rmw_connext_cpp/src/functions.cpp#L269

https://github.com/ros2/rmw_connext/blob/master/rmw_connext_cpp/src/functions.cpp#L599

It looks to me like the existing behavior is the same for readers and writers, and I need to change my initialization of the service/client readers and writers to match. I'm still not sure why the "ensure history depth is at least the requested size" block is needed though; why isn't the first part (lines 265-267) sufficient?

rmw_connext_cpp/src/functions.cpp
jacquelinekay: Alright, I'll make the necessary changes here and to rmw and rmw_opensplice. I'll leave the work to propagating the QoS profile to rclcpp until after this set gets merged, if that's ok.

esteve: No problem, we can split work and I can deal with `rclcpp` if you're ok with that.

dirk-thomas:
I was not referring to any difference between `create_publisher` and `create_subscription`. I was referring to the logic how the history depth of the data reader and writer is set.
For one value the code overrides the configuration from the xml - for the other the configuration can override the value specified in the code. I think we need to unify that and come up with a clear documentation / rational why we do it that way.

wjwwood:
Is this discrepancy covered by https://github.com/ros2/rmw/issues/13?

esteve:
I've rebased on top of master and ran the tests in `gtest-parameters` several times on Linux and it seems any issues are now gone, clients are no longer stuck waiting for a response from servers.

jacquelinekay:
Regarding the discussion about how depth is set: can we table the discussion for now and merge these PRs, since they fix several issues with parameters and services?

rosidl_typesupport_connext_cpp/resource/srv__type_support.cpp.template
```diff
@@ -95,21 +156,73 @@ int64_t send_request__@(spec.srv_name)(
...
 
   DDSDomainParticipant * participant = static_cast<DDSDomainParticipant *>(untyped_participant);
-  const DDS_DataReaderQos * datareader_qos = static_cast<const DDS_DataReaderQos *>(untyped_datareader_qos);
-  const DDS_DataWriterQos * datawriter_qos = static_cast<const DDS_DataWriterQos *>(untyped_datawriter_qos);
+  const rmw_qos_profile_t * replier_qos = static_cast<const rmw_qos_profile_t *>(untyped_replier_qos);
+  DDS_DataReaderQos datareader_qos;
+  DDS_DataWriterQos datawriter_qos;
+
+  if (!get_datareader_qos(participant, datareader_qos)) {
+    RMW_SET_ERROR_MSG("Unable to initialize datareader QoS");
```
dirk-thomas: It looks like these error messages are being overwritten afterwards if the function return NULL.

wjwwood:
I think it needs to be unified one way or the other before merging this, but I don't care which way it is unified. The discussion of which way they both should be can be tabled.

rosidl_typesupport_connext_cpp/src/dds_utilities.cpp
```diff
@@ -0,0 +1,70 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <rosidl_typesupport_connext_cpp/dds_utilities.hpp>
+
+bool
+get_datareader_qos(DDSDomainParticipant * participant, DDS_DataReaderQos & datareader_qos)
```
dirk-thomas: Moving / copying these functions from rmw into the rosidl package looks weird to me. It implies that every compiled message package embeds this logic. I would prefer if this stays in rmw and if the typesupport needs its somehow gets its passed as an argument.

wjwwood: This is compiled into the library for this package, not into each message library:

https://github.com/ros2/rmw_connext/blob/fix_service_qos/rosidl_typesupport_connext_cpp/CMakeLists.txt#L33

wjwwood: It may be true that it should be in rmw_connext_cpp instead though. The question is where is it used from. If it is used from the message specific code, then it would need to be here or as you suggested moved to rmw and used there, passing the result to the message specific code.

dirk-thomas: Good point - I didn't notice that. Before the type support library only contained the identifier.

Without this change the reader / write qos was determined within rmw so it should be possible to keep it there.

jacquelinekay: These functions are used in `srv__type_support.cpp.template, line 71:
https://github.com/ros2/rmw_connext/pull/100/files#diff-a2cb81bcab4593a61a1d8ad7a7308f25R71

I could change this function to take a DDSDataWriterQos struct and a DDSDataReaderQos struct, instead of an rmw_qos_profile, and initialize the DDS QoS structs in the rmw implementation before `create_requester` is called. Then the get_datareader/writer_qos functions would not need to be in the typesupport library.

dirk-thomas: That sounds like a good way to go.

jacquelinekay: Done. The diff is much simpler now. Not sure what I was doing before; I think I was trying to make this implementation match the OpenSplice typesupport library.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -929,11 +931,78 @@ rmw_create_client(
...
+  )
+  {
+    if (qos_profile.depth > (std::numeric_limits<DDS_Long>::max)()) {
+      RMW_SET_ERROR_MSG(
+        "failed to set history depth since the requested queue size exceeds the DDS type");
+      goto fail;
+    }
+    datareader_qos.history.depth = static_cast<DDS_Long>(qos_profile.depth);
+  }
+
```
dirk-thomas: Wouldn't it make sense to move all this code into the `get_datareader_qos` / `get_datawriter_qos` functions where all the other qos properties are set? That would also avoid have the same ~60 lines in four locations.

jacquelinekay: True, that would be a big improvement. Done in 65a3d43

rmw_connext_shared_cpp/src/shared_functions.cpp
```diff
@@ -186,6 +241,53 @@ get_datawriter_qos(DDSDomainParticipant * participant, DDS_DataWriterQos & dataw
...
+  )
+  {
+    if (qos_profile.depth > (std::numeric_limits<DDS_Long>::max)()) {
+      RMW_SET_ERROR_MSG(
+        "failed to set history depth since the requested queue size exceeds the DDS type");
+      return false;
+    }
+    datawriter_qos.history.depth = static_cast<DDS_Long>(qos_profile.depth);
+  }
+
```
dirk-thomas: `DDS_DataReaderQos` and `DDS_DataWriterQos` don't share a base class but you could implement this ~55 lines function once as a template in the cpp file and avoid having it twice.

jacquelinekay: Done. Sharing this code also forces the DataReader and DataWriter to interpret QoS profiles in the same way, which is probably what we want.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -2261,11 +2171,11 @@ rmw_create_client(
 
   // create requester
   {
-    if (!get_datareader_qos(participant, datareader_qos)) {
+    if (!get_datareader_qos(qos_profile, participant, datareader_qos)) {
```
dirk-thomas: I would pass `qos_profile` as a second or third argument here (as well as below). Not really sure why - most other functions take the participant as a first argument.

dirk-thomas:
LGTM. Let's spin some CI jobs.

jacquelinekay:
http://ci.ros2.org/job/ros2_batch_ci_linux/372/
http://ci.ros2.org/job/ros2_batch_ci_osx/278/
http://ci.ros2.org/job/ros2_batch_ci_windows/421/

dirk-thomas:
+1

PR #101, opened 2015-09-23
Use transient QoS for the requester/replier data writers/readers

esteve:
This PR adds the logic needed to apply the durability policy to data readers and writers.

Connects to ros2/rmw#41


dirk-thomas:
Instead of forcing the value to be transient local couldn't we change the default profile for service use the system default? Then if a user passes a custom profile it is still possible to set arbitrary qos values.

jacquelinekay:
I agree with Dirk; I think it's worthwhile to expose durability QoS setting and assign TRANSIENT_LOCAL as the default for client/service DataReaders/Writers, since we've discovered it is relevant to the performance of clients/services.

dirk-thomas:
Currentlythis forces the value to be `DDS_VOLATILE_DURABILITY_QOS`. Based on the above comments I thought we make this conditional based on the passed in policy?

esteve:
@dirk-thomas the last commit was part of a test did, but was pushed by error, I'll re-run the CI jobs to make sure everything is ok.

dirk-thomas:
+1

PR #102, opened 2015-10-16
update code to pass ament_cpplint

dirk-thomas:


rmw_connext_cpp/src/functions.cpp
```diff
@@ -31,25 +31,25 @@
 #  pragma clang diagnostic ignored "-Wreturn-type-c-linkage"
 # endif
 #endif
-#include <ndds/ndds_cpp.h>
-#include <ndds/ndds_requestreply_cpp.h>
+#include "ndds/ndds_cpp.h"
+#include "ndds/ndds_requestreply_cpp.h"
```
wjwwood: I don't think these count as our headers, and so I believe they should be included with `<>` and should be treated something like boost would be. I think they should fall under "Other libraries' .h files." as described in https://google-styleguide.googlecode.com/svn/trunk/cppguide.html#Names_and_Order_of_Includes. To be fair, the guide doesn't clearly state which include style should be used for "Other libraries' .h files.".

This might be a case which isn't covered by the linter and in which case we need to clearly state what we do and make the linter agree with a patch. In my opinion it should be something like this example:

(`project/src/foo.cpp`)
```c++
#include "project/foo.hpp"

#include <stdio.h>

#include <iostream>

#include <boost/thread.hpp>

#include "project/other.hpp"
```

Thoughts?

rmw_connext_cpp/src/functions.cpp
```diff
@@ -63,7 +63,6 @@ _create_type_name(
 
 extern "C"
 {
-
```
wjwwood: Why?

dirk-thomas: Because cpplint warns about it otherwise.

We could disable the rule `whitespace/blank_line` if we want to. But I am ok with following the style guide as-is.

rmw_connext_cpp/src/functions.cpp
dirk-thomas: Afaik it is recommended to use `<>` for system headers. I wouldn't classify Connext as such.

rmw_connext_cpp/src/functions.cpp
wjwwood: Seems a bit strict to me. I don't mind minimizing whitespace, i.e. warning when more than one line is between items, but this kind of sucks.

rmw_connext_cpp/src/functions.cpp
wjwwood: Hmm, I'm not sure I agree. Neither Boost nor Qt examples suggest to user to use quotes when including their headers and they are not "system headers". The Google style guide doesn't say one way or the other and neither does the "include what you use" tool from Google. I don't see good rationals out there for one way or the other, but this just seems like using quotes to make the linter happy.

I suppose we just forget about it.

dirk-thomas: For these lines cpplint actually doesn't care. I reverted them to how they were before the PR.

wjwwood:
+1

PR #108, opened 2015-11-05
additional cmake logic to allow disabling one connext implementation or another

wjwwood:
Fixes ros2/ros2#147

### DIFF
--- a/rmw_connext_cpp/CMakeLists.txt
+++ b/rmw_connext_cpp/CMakeLists.txt
@@ -2,6 +2,9 @@ cmake_minimum_required(VERSION 2.8.3)
 
 project(rmw_connext_cpp)
 
+set(CONNEXT_STATIC_DISABLE $ENV{CONNEXT_STATIC_DISABLE}
+  CACHE BOOL "If Connext Static should be disabled.")
+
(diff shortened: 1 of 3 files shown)


dirk-thomas:
Does this allow you to reenable one rmw impl. using the environment variable after having it disabled before?

wjwwood:
No, the logic to override the cache value with the environment was more complicated. I guess I could implement that, but I think it would be better to use the CMake cache variables at that point. The only reason I think we must have the the environment variables is so we can affect the farm without passing unused variables to all the other cmake packages (producing lots of unnecessary cmake warnings). These shouldn't be a problem when working locally, so I think the CMake variables should be used in that case.

dirk-thomas:
Do we need the changes to `rmw_connext_shared_cpp`? Until now that package didn't need to know about which packages are using it. E.g. if someone adds a third rmw implementation this logic breaks. If someone wants to disables both he should just disabled Connext alltoget (e.g. not source Connext in the first place).

wjwwood:
No, we can probably leave it out of this, but I was trying to avoid as much work as possible.

dirk-thomas:
+1

PR #109, opened 2015-11-05
Fix ddsgen server

dirk-thomas:
Doesn't have any effect on the farm where the tool is actually in /usr/bin so no CI job.


esteve:
+1

PR #110, opened 2015-11-05
update connext environment generation

dirk-thomas:
With this patch it is sufficient to source the environment of the ament workspace which will implicitly set all variables necessary to use Connext. Before Connext still needed to be source explicitly (at least on non-Windows platforms).

On Linux the library dir doesn't need to be on the path anymore.

On Windows the environment hook sources the "right" batch file based on the architecture name.

In both cases it warns now if the NDDSHOME changes.

http://ci.ros2.org/job/ros2_batch_ci_linux/554/
http://ci.ros2.org/job/ros2_batch_ci_osx/422/
http://ci.ros2.org/job/ros2_batch_ci_windows/569/


tfoote:
+1

PR #113, opened 2015-12-02
Implement waitset handle

jacquelinekay:
Connects to ros2/rmw#49

### DIFF
--- a/rmw_connext_shared_cpp/src/shared_functions.cpp
+++ b/rmw_connext_shared_cpp/src/shared_functions.cpp
@@ -510,6 +510,189 @@ destroy_guard_condition(const char * implementation_identifier,
   return result;
 }
 
+rmw_waitset_t *
+create_waitset(const char * implementation_identifier,
+  rmw_guard_conditions_t * fixed_guard_conditions, size_t max_conditions)
+{
+  rmw_waitset_t * waitset = rmw_waitset_allocate();
+
+  ConnextWaitSetInfo * waitset_info = nullptr;
+  DDSGuardCondition * dds_guard_cond = nullptr;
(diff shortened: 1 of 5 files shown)


jacquelinekay:
here's something pretty cool:

when I run the allocator test with --all-tokens on the waitset_handle set of branches (meaning no functions are ignored from the middleware), the connext version of the test passes!

rmw_connext_shared_cpp/src/shared_functions.cpp
```diff
@@ -510,6 +510,185 @@ destroy_guard_condition(const char * implementation_identifier,
...
+  rmw_waitset_t * waitset = rmw_waitset_allocate();
+
+  ConnextWaitSetInfo * waitset_info = nullptr;
+  DDSGuardCondition * dds_guard_cond = nullptr;
+  DDS_ReturnCode_t ret;
+
+  // From here onward, error results in unrolling in the goto fail block.
+  if (!waitset) {
+    RMW_SET_ERROR_MSG("failed to allocate waitset");
+    goto fail;
```
dirk-thomas: In this case `waitset_info` is a `nullptr` but in the first line under `fail` `waitset_info->active_conditions` is being accessed.

jacquelinekay: 757002b8282f733a93f074c8352f059ed7327c8b will fix this by checking the pointer in the fail block.

jacquelinekay:
I believe this and ros2/rmw_opensplice#98 are the last pull requests that need approval and that I have addressed all comments. @dirk-thomas what do you think?

dirk-thomas:
lgtm.

PR #114, opened 2015-12-04
add missing run dependency on rmw_connext_shared_cpp

dirk-thomas:
Same as https://github.com/ros2/rmw_connext/blob/43f847c3545d400fd995a1370579e951a3c930b1/rmw_connext_cpp/package.xml#L28


esteve:
+1

PR #115, opened 2015-12-07
add missing dependency on rmw

dirk-thomas:


dirk-thomas:
Nvw. this patch contains too many thing. I will split it out.

PR #118, opened 2015-12-15
changes to support rcl refactor

wjwwood:
Connects to ros2/rcl#5

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -150,6 +150,11 @@ rmw_create_publisher(
     rosidl_typesupport_connext_cpp::typesupport_connext_identifier,
     return NULL)
 
+  if (!qos_profile) {
+    RMW_SET_ERROR_MSG("qos_profile is null");
+    return nullptr;
+  }
+
   auto node_info = static_cast<ConnextNodeInfo *>(node->data);
   if (!node_info) {
     RMW_SET_ERROR_MSG("node info handle is null");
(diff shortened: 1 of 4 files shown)


rmw_connext_cpp/src/functions.cpp
```diff
@@ -230,7 +230,7 @@ rmw_create_publisher(
     }
   }
 
-  if (!get_datawriter_qos(participant, qos_profile, datawriter_qos)) {
+  if (!get_datawriter_qos(participant, *qos_profile, datawriter_qos)) {
```
dirk-thomas: Should the Connext shared API be updated to use a pointer rather then a reference?

wjwwood: It could I guess, but since it isn't pure C it doesn't have to. I was just going for minimal changes.

jacquelinekay: 2 things:

I think it's fine if the shared API keeps a reference to the qos_profile, because the shared API is C++. That way we can avoid duplicating code to check that the pointer is not null before dereferencing. On that note, do you need to check if qos_profile is null here before dereferencing it?

rmw_connext_cpp/src/functions.cpp
```diff
@@ -514,7 +514,7 @@ rmw_create_subscription(const rmw_node_t * node,
     }
   }
 
-  if (!get_datareader_qos(participant, qos_profile, datareader_qos)) {
+  if (!get_datareader_qos(participant, *qos_profile, datareader_qos)) {
```
jacquelinekay: Same comment here as above.

dirk-thomas:
+1

wjwwood:
I had to add another change here to get the node tests to pass. Please have a look at 32b7b52.

rmw_connext_cpp/src/functions.cpp
wjwwood: Yup, good point a580140. I'll update OpenSplice as well.

rmw_connext_cpp/src/functions.cpp
wjwwood: a580140

PR #119, opened 2015-12-17
adding missing dependencies

tfoote:
on rosidl_typesupport_introspection_c necessary for ros2/rclcpp#182

Connects to  ros2/rclcpp#181

### DIFF
--- a/rmw_connext_dynamic_cpp/package.xml
+++ b/rmw_connext_dynamic_cpp/package.xml
@@ -13,11 +13,13 @@
   <build_depend>rmw</build_depend>
   <build_depend>rmw_connext_shared_cpp</build_depend>
   <build_depend>rosidl_generator_cpp</build_depend>
+  <build_depend>rosidl_typesupport_introspection_c</build_depend>
   <build_depend>rosidl_typesupport_introspection_cpp</build_depend>
 
   <build_export_depend>connext_cmake_module</build_export_depend>
   <build_export_depend>libndds51</build_export_depend>
   <build_export_depend>rosidl_generator_cpp</build_export_depend>
+  <build_export_depend>rosidl_typesupport_introspection_c</build_export_depend>
   <build_export_depend>rosidl_typesupport_introspection_cpp</build_export_depend>
 
   <exec_depend>rmw</exec_depend>


dirk-thomas:
Can you please explain the dependency chain.

tfoote:
Without this dependency the rosidl_typesupport_introspection_c does not get sourced in the test environment setup.  `cmake__test.sh`

This is where @wjwwood and I found a dependency on cpp without a matching c dependency. Since the cpp was in the environment but not the c.

dirk-thomas:
While this might make it work for your case we need to know *why* this is necessary. That is why I asked for the chain: which part A requires something from `rosidl_typesupport_introspection_c`. Without that knowledge this is only fixing the immediate symptom.

wjwwood:
I agree that these are necessary, because without them `rosidl_typesupport_introspection_c` doesn't get it's `local_setup.sh` sourced when running tests in `rclcpp`.  I think it is correct to say that any rmw implementation that depends on `rosidl_typesupport_introspection_cpp` should also depend on `rosidl_typesupport_introspection_c`.

It could be a different argument that the rmw implementations which use those typesupport introspection libraries shouldn't depend on them, but I can't tell if that's true or not. Either way this fixes the problem that both @tfoote and I were running into on OS X.

tfoote:
The missing library was referenced as follows: 

```
4: Test timeout computed to be: 60
4: -- run_test.py: invoking following command in '/Users/osrf/tfoote/ros2_ws/build_isolated/rclcpp':
4:  - /Users/osrf/tfoote/ros2_ws/build_isolated/rclcpp/test_rate --gtest_output=xml:/Users/osrf/tfoote/ros2_ws/build_isolated/rclcpp/test_results/rclcpp/test_rate.gtest.xml
4: dyld: Library not loaded: @rpath/librosidl_typesupport_introspection_c.dylib
4:   Referenced from: /Users/osrf/tfoote/ros2_ws/install_isolated/rcl_interfaces/lib/librcl_interfaces__rosidl_typesupport_introspection_c.dylib
```

dirk-thomas:
@tfoote The question is not if it was referenced. The question it why?

Since Connext does not directly use anything from `rosidl_typesupport_introspection_c` I don't think this is the right place to fix it.

tfoote:
@dirk-thomas @wjwwood asked me to post that to help isolate where it's coming from

tfoote:
fixing with rosidl_default_runtime package

PR #122, opened 2016-02-02
Connext C typesupport

jacquelinekay:
In progress (there are no tests).

Mostly a search/replace of an early version of #106. I may need to update it to get feature parity with the new version of Opensplice C typesupport, need to take a look at it first.

Needs ros2/rosidl#79

### DIFF
--- a/rosidl_typesupport_connext_c/CMakeLists.txt
+++ b/rosidl_typesupport_connext_c/CMakeLists.txt
@@ -1,0 +1,73 @@
+cmake_minimum_required(VERSION 2.8.3)
+
+project(rosidl_typesupport_connext_c)
+
+if(NOT WIN32)
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
+endif()
+
+find_package(ament_cmake REQUIRED)
+
+find_package(connext_cmake_module REQUIRED)
+find_package(Connext MODULE)
+if(NOT Connext_FOUND)
+  ament_package()
+  message(WARNING "Could not find RTI Connext - skipping '${PROJECT_NAME}'")
+  return()
+endif()
+
+find_package(ament_cmake_python REQUIRED)
+find_package(rosidl_typesupport_connext_cpp REQUIRED)
+
+ament_export_dependencies(rmw)
+ament_export_dependencies(rosidl_cmake)
+ament_export_dependencies(rosidl_generator_c)
+ament_export_dependencies(rosidl_generator_dds_idl)
+ament_export_dependencies(rosidl_typesupport_connext_cpp)
+# The reason the impl folder is exported is that it contains the implementation
+# for the get_*_type_support_handle functions and defines the connext
+# specific version of these functions.
+ament_export_include_directories(include include/${PROJECT_NAME}/impl)
+
+ament_python_install_package(${PROJECT_NAME})
+
+add_library(${PROJECT_NAME} SHARED src/identifier.cpp)
+if(WIN32)
+  target_compile_definitions(${PROJECT_NAME}
+    PRIVATE "ROSIDL_TYPESUPPORT_CONNEXT_C_BUILDING_DLL")
+endif()
+target_include_directories(${PROJECT_NAME} PUBLIC include)
+ament_target_dependencies(${PROJECT_NAME} "rosidl_typesupport_connext_cpp")
+ament_export_libraries(${PROJECT_NAME})
+
+if(AMENT_ENABLE_TESTING)
+  find_package(ament_lint_auto REQUIRED)
+  ament_lint_auto_find_test_dependencies()
+endif()
+
+ament_package(
+  CONFIG_EXTRAS "rosidl_typesupport_connext_c-extras.cmake.in"
+)
+
+install(
+  PROGRAMS bin/rosidl_typesupport_connext_c
+  DESTINATION lib/rosidl_typesupport_connext_c
+)
+
+install(
+  DIRECTORY cmake resource
+  DESTINATION share/${PROJECT_NAME}
+)
+
+install(
+  DIRECTORY include/
+  DESTINATION include
+)
+
+install(
+  TARGETS ${PROJECT_NAME}
+  ARCHIVE DESTINATION lib
+  LIBRARY DESTINATION lib
+  RUNTIME DESTINATION bin
+)
--- a/rosidl_typesupport_connext_c/bin/rosidl_typesupport_connext_c
+++ b/rosidl_typesupport_connext_c/bin/rosidl_typesupport_connext_c
@@ -1,0 +1,44 @@
+#!/usr/bin/env python3
+
+import argparse
+import os
+import sys
+
+from rosidl_cmake import read_generator_arguments
+from rosidl_parser import UnknownMessageType
+from rosidl_typesupport_connext_c import generate_typesupport_connext_c
+
+
+def is_valid_file(parser, file_name):
+    if not os.path.exists(file_name):
+        parser.error("File does not exist: '{0}'".format(file_name))
+    file_name_abs = os.path.abspath(file_name)
+    if not os.path.isfile(file_name_abs):
+        parser.error("Path exists but is not a file: '{0}'".format(file_name_abs))
+    return file_name
+
+
+def main(argv=sys.argv[1:]):
+    parser = argparse.ArgumentParser(
+        description='Generate the C interfaces for Connext.',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+    parser.add_argument(
+        '--generator-arguments-file',
+        required=True,
+        help='The location of the file containing the generator arguments')
+    args = parser.parse_args(argv)
+
+    generator_args = read_generator_arguments(args.generator_arguments_file)
+
+    try:
+        rc = generate_typesupport_connext_c(generator_args)
+    except UnknownMessageType as e:
+        print(str(e), file=sys.stderr)
+        return 1
+    if rc:
+        return rc
+    return 0
+
+
+if __name__ == '__main__':
+    sys.exit(main())
--- a/rosidl_typesupport_connext_c/cmake/rosidl_typesupport_connext_c_generate_interfaces.cmake
+++ b/rosidl_typesupport_connext_c/cmake/rosidl_typesupport_connext_c_generate_interfaces.cmake
@@ -1,0 +1,258 @@
+# Copyright 2016 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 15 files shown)


rosidl_typesupport_connext_c/CMakeLists.txt
```diff
@@ -0,0 +1,72 @@
+cmake_minimum_required(VERSION 2.8.3)
+
+project(rosidl_typesupport_connext_c)
+
+if(NOT WIN32)
+  set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=c11 -Wall -Wextra")
```
wjwwood: Since the `identifier.cpp` file is built by this package, I think you need to set the `CMAKE_CXX_FLAGS` as well. Also I'd recommend dropping the `-std=c11` flag, since we're not using anything that is C11 specific right now.

rosidl_typesupport_connext_c/cmake/rosidl_typesupport_connext_c_generate_interfaces.cmake
```diff
@@ -0,0 +1,242 @@
...
+        PROPERTIES COMPILE_FLAGS ${_connext_compile_flags})
+    endforeach()
+  endif()
+endif()
+
+set(_dependency_files "")
+set(_dependencies "")
+foreach(_pkg_name ${rosidl_generate_interfaces_DEPENDENCY_PACKAGE_NAMES})
+  foreach(_idl_file ${${_pkg_name}_INTERFACE_FILES})
+  get_filename_component(_extension "${_idl_file}" EXT)
```
wjwwood: This indention looks weird to me, is it right?

jacquelinekay: It is not correct. Wonder why the linter didn't catch that.

rosidl_typesupport_connext_c/bin/rosidl_typesupport_connext_c
```diff
@@ -0,0 +1,52 @@
...
+        required=True,
+        help='The location of the file containing the generator arguments')
+    parser.add_argument(
+        '--dds-interface-base-path',
+        required=True,
+        help='The base location of the DDS interface files')
+    parser.add_argument(
+        '--idl-pp',
+        required=True,
+        help='The location of the IDL preprocessor')
```
wjwwood: I realized that we can remove these two options from the script because they are only used by the part we delegate to the C++ generator. Here is the similar change I just made for opensplice: https://github.com/ros2/rmw_opensplice/commit/893bb5baf7b66a440bfe04b1dbee66a5fbc0b327

rosidl_typesupport_connext_c/cmake/rosidl_typesupport_connext_c_generate_interfaces.cmake
```diff
@@ -0,0 +1,242 @@
...
+  "rosidl_typesupport_connext_c")
+
+add_dependencies(
+  ${rosidl_generate_interfaces_TARGET}
+  ${rosidl_generate_interfaces_TARGET}${_target_suffix}
+)
+add_dependencies(
+  ${rosidl_generate_interfaces_TARGET}${_target_suffix}
+  ${rosidl_generate_interfaces_TARGET}__cpp
+)
```
wjwwood: I think you may need to add another dependency on the DDS IDL file generation which is done in the C++ generator. The target is created here: https://github.com/ros2/rmw_connext/blob/master/rosidl_typesupport_connext_cpp/cmake/rosidl_typesupport_connext_cpp_generate_interfaces.cmake#L25

jacquelinekay: I actually had an `add_dependencies` line for that target originally and it didn't compile for some reason. Maybe there was something I missed the first time around.

https://github.com/ros2/rmw_connext/commit/8e48d6364d0e8d30fb3887a3b7fc9ead238942f3

rosidl_typesupport_connext_c/include/rosidl_typesupport_connext_c/message_type_support.h
```diff
@@ -0,0 +1,40 @@
...
+  bool (* take)(
+    void * dds_data_reader, bool ignore_local_publications, void * ros_message, bool * taken,
+    void * sending_publication_handle);
+  const char * (*convert_ros_to_dds)(
+    const void * untyped_ros_message,
+    void * untyped_data_message);
+  const char * (*convert_dds_to_ros)(
+    const void * untyped_data_message,
+    void * untyped_ros_message);
+} message_type_support_callbacks_t;
```
wjwwood: This structure must be the same as the one in `rosidl_typesupport_connext_cpp` (defined here: https://github.com/ros2/rmw_connext/blob/master/rosidl_typesupport_connext_cpp/include/rosidl_typesupport_connext_cpp/message_type_support.h#L20-L32), other wise the `rmw_connext_cpp` code will cast it from a `void *` to the wrong type. Because of this I think you should change the one in `rosidl_typesupport_connext_cpp` to include the convert function pointers and include that header here.

rosidl_typesupport_connext_c/resource/msg__type_support_c.cpp.template
```diff
@@ -0,0 +1,539 @@
...
+  if (!untyped_ros_message) {
+    return "ros message handle is null";
+  }
+  if (!untyped_dds_message) {
+    return "dds message handle is null";
+  }
+  const __ros_msg_type * ros_message = static_cast<const __ros_msg_type *>(untyped_ros_message);
+  __dds_msg_type * dds_message = static_cast<__dds_msg_type *>(untyped_dds_message);
+  (void)dds_message;
+  (void)ros_message;
```
wjwwood: I moved these in the OpenSplice code.

rosidl_typesupport_connext_c/rosidl_typesupport_connext_c/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+from rosidl_cmake import get_newest_modification_time
+from rosidl_parser import parse_message_file
+from rosidl_parser import parse_service_file
+from rosidl_parser import validate_field_types
+
+
+def generate_typesupport_connext_c(args):
+    template_dir = args['template_dir']
+    mapping_msgs = {
+        # os.path.join(template_dir, 'msg__type_support.hpp.template'): '%s__type_support.hpp',
```
wjwwood: I removed this in the opensplice code.

wjwwood:
Other than my comments, it looks good to me. Of course any changes I do to opensplice will probably need to be reflected here (it just reminds me how much code duplication we have in the type support code...)

rosidl_typesupport_connext_c/include/rosidl_typesupport_connext_c/message_type_support.h
jacquelinekay: Changed in f39c308

jacquelinekay:
http://ci.ros2.org/job/ci_linux/880
http://ci.ros2.org/job/ci_osx/737
http://ci.ros2.org/job/ci_windows/964

jacquelinekay:
investigating problem with missing installed headers

jacquelinekay:
Can't reproduce the problem seen on Jenkins locally after nuking my workspace temporary files and trying both an isolated and non-isolated build. Argh,

jacquelinekay:
I think @wjwwood was correct and the missing `add_dependency` will fix the build on Jenkins. Still not sure why I couldn't reproduce that failure locally.

http://ci.ros2.org/job/ci_linux/884/
http://ci.ros2.org/job/ci_osx/740/
http://ci.ros2.org/job/ci_windows/968/

jacquelinekay:
Now I need to address some warnings with Windows.

jacquelinekay:
I guess the warnings are from console_bridge, actually. I'll investigate what's going on with the failing services test on this build.

jacquelinekay:
I think the failing tests on Linux are a fluke; they pass locally and the console output from gtest from the Jenkins build is OK:

http://ci.ros2.org/job/ci_linux/884/testReport/junit/projectroot.home.rosbuild.ci_scripts.ws.src.ros2.system_tests/test_rclcpp/test_services_cpp__rmw_connext_dynamic_cpp/

Therefore I think this is OK to merge.

jacquelinekay:
Hmm, testing locally with a modified version of William's c_typesupport_tests, I'm getting link errors in the rcl tests for rmw_fastrtps_cpp and rmw_connext_dynamic_cpp (I'm using rmw_implementation, instead of the Opensplice-only hack). Will take a look.

jacquelinekay:
Alright, after properly configuring the `c_typesupport_test` branches by @wjwwood, I realized that I wasn't properly building in linking in the right symbols for this PR. Retriggering CI, although I suspect Windows isn't going to build.

http://ci.ros2.org/job/ci_linux/903/
http://ci.ros2.org/job/ci_osx/754/
http://ci.ros2.org/job/ci_windows/981/

jacquelinekay:
hopefully getting rid of OSX warnings by removing redundant sources in the makefile:

http://ci.ros2.org/job/ci_osx/761

jacquelinekay:
The warnings in OSX are fixed. Now trying to fix Windows...

jacquelinekay:
Alright! The latest Windows build is clean except for warnings in the generated code which will be fixed by a9df4a4. I can run another job if others need to be convinced.

http://ci.ros2.org/job/ci_windows/1016/

jacquelinekay:
one last job with Connext only to see if the warnings were fixed, then I'll squash and merge

http://ci.ros2.org/job/ci_windows/

jacquelinekay:
well, the Connext-only option is broken (build failed because it couldn't find opensplice_cmake_module), so I'm restarting Jenkins.

dirk-thomas:
Is this only the case for your branch? I thought master was supposed to work when only specific rmw implementations are selected.

jacquelinekay:
I've queued a job testing master with FastRTPS and Opensplice disabled: http://ci.ros2.org/job/ci_windows/1029

dirk-thomas:
The Windows job failed. But this is a use case which is supposed to work. I am not sure which recent change broke it but it needs to be fixed again.

jacquelinekay:
indeed, selecting middlewares on Jenkins is broken. I guess we don't have automated tests for our CI options.

to clarify, the most recent job on all middleware impls for this particular branch was successful, which is why I merged this branch. I also put in a fix before merging that should fix the warnings in the generated code.
http://ci.ros2.org/job/ci_windows/1028/

rosidl_typesupport_connext_c/cmake/rosidl_typesupport_connext_c_generate_interfaces.cmake
```diff
@@ -0,0 +1,258 @@
...
+  ROS_INTERFACE_FILES "${rosidl_generate_interfaces_IDL_FILES}"
+  ROS_INTERFACE_DEPENDENCIES "${_dependencies}"
+  OUTPUT_DIR "${_output_path}"
+  TEMPLATE_DIR "${rosidl_typesupport_connext_c_TEMPLATE_DIR}"
+  TARGET_DEPENDENCIES ${target_dependencies}
+  ADDITIONAL_FILES ${_dds_idl_files}
+)
+
+add_custom_command(
+  OUTPUT ${_generated_msg_files} ${_generated_srv_files}
```
dirk-thomas: This line results in a CMake error if a package has only service files.

dirk-thomas: See #135.

PR #123, opened 2016-02-06
export optimized as well as debug libraries if found

### LINKED ISSUE #86: Setup windows debug libraries in Connext_LIBRARIES (cmake module)

dirk-thomas:
Fixes #86.

Before: http://ci.ros2.org/job/ci_windows/994/
After: http://ci.ros2.org/job/ci_windows/997/

### DIFF
--- a/connext_cmake_module/cmake/Modules/FindConnext.cmake
+++ b/connext_cmake_module/cmake/Modules/FindConnext.cmake
@@ -48,42 +48,133 @@
 
 set(Connext_FOUND FALSE)
 
-# check if all libraries with an expected name have been found
-function(_find_connext_ensure_libraries var expected_library_names library_paths)
-  foreach(expected_library_name ${expected_library_names})
-    set(found FALSE)
-    foreach(library_path ${library_paths})
-      get_filename_component(library_name "${library_path}" NAME)
-      if("${expected_library_name}" STREQUAL "${library_name}")
-        set(found TRUE)
-        break()
-      endif()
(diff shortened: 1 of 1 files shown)


wjwwood:
Reading over the code looks ok, but I have some major conflicts locally so I didn't run it on my mac. I'm happy with the changes if CI is. The Windows CI is probably enough, but do you intend to run CI for Linux/OS X?

dirk-thomas:
I ran all platform with both build types successfully. Since only Windows has a relevant change (from didn't work before to works now) I didn't link them all.

PR #126, opened 2016-02-10
use imported cmake targets for Connext libraries

wjwwood:
By using imported CMake targets for the Connext libraries, it allows CMake to automatically inject the location of these libraries into the `RPATH` of libraries we link them against. This is part of what is needed to run things linked against Connext on OS X without `DYLD_*` environment variables.

Without these changes a library linked against one of the Connext libraries would not have an `RPATH` entry with the directory where the Connext library was found and so the ID of the Connext library, e.g. `@rpath/libnddsc.dylib`, would not resolve at run time.

I considered adding the `Connext_LIBRARY_DIR` as an explicit `RPATH` on each of our libraries that we link to `Connext_LIBRARIES`, but this is a far less intrusive change.

I've taken care to preserve the different libraries being passed along for different build configurations, but I'm unsure how to test that it actually works. @dirk-thomas can you check that out or tell me how to do that?

References:
- https://cmake.org/Wiki/CMake_RPATH_handling
- https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man1/dyld.1.html
- http://matthew-brett.github.io/docosx/mac_runtime_link.html
- https://cmake.org/pipermail/cmake/2014-November/059156.html

CI (Connext only):
- http://ci.ros2.org/job/ci_linux/918/
- http://ci.ros2.org/job/ci_osx/779/
- http://ci.ros2.org/job/ci_windows/1020/


connext_cmake_module/cmake/Modules/FindConnext.cmake
```diff
@@ -211,29 +209,38 @@ if(NOT "${_NDDSHOME} " STREQUAL " ")
...
-      # if libraries of both types are found use prepend the build configuration
-      if(_found_all_debug_libraries)
-        list(APPEND Connext_LIBRARIES "optimized")
-      endif()
-      list(APPEND Connext_LIBRARIES "${_lib_path}/${_lib}")
+    # for each of the optimized libraries set the IMPORTED_LOCATION for the RELEASE config
+    foreach(_val RANGE ${_loop_iterations})
+      list(GET _base_names ${_val} _base_name)
+      list(GET _optimized_libraries ${_val} _lib)
+      set_target_properties(${_base_name} PROPERTIES IMPORTED_LOCATION_RELEASE "${_lib_path}/${_lib}")
```
dirk-thomas: What about the location for other release-like configurations like `RelWithDebInfo`?

dirk-thomas:
To test this you need a Connext version which provides both library types. Then you need to build the workspace with different build types and confirm that it links against the expected libraries in each case.

dirk-thomas:
You probably also want to test the case where Connext has only one kind of libraries to ensure the CMake code works with that too.

wjwwood:
Well, my machine has both (with and without the `d` suffix):

```
% ls /Applications/rti_connext_dds-5.2.0/lib/x64Darwin14clang6.0
libnddsc.dylib                libnddscppd.dylib             librticonnextmsgc.dylib       librtidlcppd.dylib            librtiroutingservicez.a
libnddscd.dylib               libnddscppz.a                 librticonnextmsgcd.dylib      librtidlcppz.a                librtiroutingservicezd.a
libnddscore.dylib             libnddscppzd.a                librticonnextmsgcpp.dylib     librtidlcppzd.a               librtirsassigntransf.dylib
libnddscored.dylib            libnddscz.a                   librticonnextmsgcppd.dylib    librtidlcz.a                  librtirsinfrastructure.dylib
libnddscorez.a                libnddsczd.a                  librticonnextmsgcppz.a        librtidlczd.a                 librtirsinfrastructured.dylib
libnddscorezd.a               libnddsjava.dylib             librticonnextmsgcppzd.a       librtimonitoring.dylib        librtirsinfrastructurez.a
libnddscpp.dylib              libnddsjavad.dylib            librticonnextmsgcz.a          librtimonitoringd.dylib       librtirsinfrastructurezd.a
libnddscpp2.dylib             libnddstransporttcp.dylib     librticonnextmsgczd.a         librtimonitoringz.a           librtirsjniadapter.dylib
libnddscpp2d.dylib            libnddstransporttcpd.dylib    librtidlc.dylib               librtimonitoringzd.a
libnddscpp2z.a                libnddstransporttcpz.a        librtidlcd.dylib              librtiroutingservice.dylib
libnddscpp2zd.a               libnddstransporttcpzd.a       librtidlcpp.dylib             librtiroutingserviced.dylib
```

And I always build with `RelWithDebInfo`. I believe it defaults to `_RELEASE` as the config.

I'll try `Debug`, `Release`, and `None` as I have time.

connext_cmake_module/cmake/Modules/FindConnext.cmake
wjwwood: Actually digging into it, I think we should use something like this: http://stackoverflow.com/a/24403456/671658

wjwwood:
I aliased `RelWithDebInfo` and `MinSizeRel` to `Release`.

I also had to rework the logic to allow for repeated invocations of the module. Since this is uncommonly desired, and probably costs us some time, I also added a resettable include guard.

dirk-thomas:
Instead of the custom logic for resetting please use something more common. If the `FOUND` flag is set the module could decide not to run again. And by resetting it the user can force it to search again.

wjwwood:
Yeah, that's simpler.

connext_cmake_module/cmake/Modules/FindConnext.cmake
```diff
@@ -42,21 +42,21 @@
 #   find_package(Connext MODULE)
 #   # use Connext_* variables
 #
+# In order to make repeated calls to find_package(Connext ...) after a success
+# then you must first reset the Connext_FOUND variable to False.
```
dirk-thomas: I would remove the `then`? As a non-english speaker though :wink:

dirk-thomas:
Does the else block from #127 also need imported targets or do you expect this never to be the case on OSX?

wjwwood:
I don't know, probably it can be excluded, but there does seem to be an opportunity for reducing the amount of code doing the same thing if both of them create imported targets. I probably won't have time to look into it before the alpha.

dirk-thomas:
Beside the last comment about the comment this lgtm.

mikaelarguedas:
@wjwwood

wjwwood:
So this would need to be update for two cases:

- The `else` part re-added in this pr: https://github.com/ros2/rmw_connext/pull/127
- When using multiple build types: https://github.com/ros2/rmw_connext/pull/126#r52530807

If I were to do that and merge this, it would only be a small part of working without setting `DYLD_LIBRARY_PATH`, which means using `RPATH` in all the shared libraries. This just fixes Connext, so it's not a full fix.

So given it's not done yet, it wouldn't be a full fix, and I don't have time for it, I'm just going to close it for now. The issue history might be useful to someone in the future if they want to get this working.

PR #127, opened 2016-02-10
readd find_package logic

dirk-thomas:
Fix regression of #123. The `else` should never have been removed: https://github.com/ros2/rmw_connext/pull/123/files#diff-f72dd578a73ca38cf3b97780449b385cL206

Before: http://ci.ros2.org/job/ci_linux/928/
After: http://ci.ros2.org/job/ci_linux/933/


connext_cmake_module/connext_cmake_module-extras.cmake
```diff
@@ -14,4 +14,5 @@
 
 # copied from connext_cmake_module/connext_cmake_module-extras.cmake
 
+find_package(ament_cmake REQUIRED)
```
wjwwood: Is this for `list_append_unique`?

dirk-thomas: Yes.

wjwwood: Hmm, that function should probably be name spaced. It would suck if someone else wrote a macro called that which worked differently from the one in ament and then tried to use it in conjunction with other ament stuff.

Obviously not something that needs to be addressed in this pr.

wjwwood:
lgtm

I'll have to update my pr too.

gerkey:
Fixed my problem using our custom packages, thanks.

PR #128, opened 2016-02-11
read CONNEXT_STATIC_DISABLE

jacquelinekay:
Left this out in the first pass for rosidl_typesupport_connext_c.

### DIFF
--- a/rmw_connext_dynamic_cpp/package.xml
+++ b/rmw_connext_dynamic_cpp/package.xml
@@ -18,6 +18,7 @@
   <build_export_depend>connext_cmake_module</build_export_depend>
   <build_export_depend>libndds51</build_export_depend>
   <build_export_depend>rosidl_generator_cpp</build_export_depend>
+  <build_export_depend>rosidl_typesupport_introspection_c</build_export_depend>
   <build_export_depend>rosidl_typesupport_introspection_cpp</build_export_depend>
 
   <exec_depend>rmw</exec_depend>
--- a/rosidl_typesupport_connext_c/CMakeLists.txt
+++ b/rosidl_typesupport_connext_c/CMakeLists.txt
@@ -2,6 +2,9 @@ cmake_minimum_required(VERSION 2.8.3)
 
 project(rosidl_typesupport_connext_c)
 
+set(CONNEXT_STATIC_DISABLE $ENV{CONNEXT_STATIC_DISABLE}
+  CACHE BOOL "If Connext Static should be disabled.")
+
 if(NOT WIN32)
   set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra")
   set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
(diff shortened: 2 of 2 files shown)


jacquelinekay:
http://ci.ros2.org/job/ci_linux/939 will test the "DISABLE_CONNEXT_STATIC" option on Jenkins, which should mean that rosidl_typesupport_c does not get built (along with rosidl_typesupport_cpp).

dirk-thomas:
This also needs a few lines more from the cpp CMake to get the variable from the environment in the first place.

jacquelinekay:
Done and rebased, retriggering:

http://ci.ros2.org/job/ci_linux/941/

I also decided to disable Opensplice and FastRTPS on this one just for the sake of speed.

jacquelinekay:
issue with the workspace. retriggered again http://ci.ros2.org/job/ci_linux/942/

tfoote:
I reran it here: http://ci.ros2.org/job/ci_linux/951/console

jacquelinekay:
The problem is that rmw_connext_dynamic_cpp exports rosidl_typesupport_introspection_c but does not include it in the package.xml or find_package it in CMakeLists.txt. I'm not sure how this dependency is resolved in the normal build, but when rosidl_typesupport_connext_c is excluded this causes the first package that depends on rmw_connext_dynamic_cpp to fail.

https://github.com/ros2/rmw_connext/blob/master/rmw_connext_dynamic_cpp/CMakeLists.txt#L40

I believe I can resolve this by either removing the export dependency or by adding the package to package.xml and find_package-ing it.

jacquelinekay:
It's not needed yet, but `rosidl_typesupport_introspection_c` will probably be needed by `rmw_connext_dynamic_cpp` once the introspection typesupport problems are solved. So I'm going to add it.

jacquelinekay:
http://ci.ros2.org/job/ci_linux/955

dirk-thomas:
If the package itself doesn't need it it shouldn't find it and shouldn't export it. If another package needs it as a transitive dependency that package is in charge to pull it in.

jacquelinekay:
Hmm, I guess it exports the dependency because it registers rosidl_typesupport_introspection_c as a possible typesupport in https://github.com/ros2/rmw_connext/blob/master/rmw_connext_dynamic_cpp/CMakeLists.txt#L44

dirk-thomas:
http://ci.ros2.org/job/ci_linux/956/ will build https://github.com/ros2/rmw_connext/tree/fix_middleware_select2 with the dependency and export  being removed.

dirk-thomas:
Also in order to export a dependency you don't need a build dependency and find_package it.

jacquelinekay:
K. I'm fine with either solution as long as it works. We will add back registering `rosidl_typesupport_introspection_c` as a typesupport impl and properly depend on it once C introspection for this package is fixed, but yours is a more correct solution for the current state of the codebase.

jacquelinekay:
One more after discussing some changes http://ci.ros2.org/job/ci_linux/957/

dirk-thomas:
It still needs: `ament_export_dependencies(rosidl_typesupport_connext_c)`

jacquelinekay:
But this is dynamic implementation, so we should not need the generated messages from rosidl_typesupport_connext_c.

dirk-thomas:
I actually meant `rosidl_typesupport_introspection_c`.

jacquelinekay:
Actually, the export dependency is already included in CMakeLists.txt on master, which is what I thought was the original cause of the problem:

https://github.com/ros2/rmw_connext/blob/master/rmw_connext_dynamic_cpp/CMakeLists.txt#L40

dirk-thomas:
+1 for the current state.

PR #129, opened 2016-02-11
Reset fixed guard conditions

### LINKED ISSUE #125: Fixed guard conditions are not reset
@dirk-thomas and I identified an apparent bug: while the non-fixed guard conditions are [reset to false after waiting](https://github.com/ros2/rmw_connext/blob/master/rmw_connext_shared_cpp/include/rmw_connext_shared_cpp/shared_functions.hpp#L406-L434), the fixed guard conditions that were [added during waitset creation](https://github.com/ros2/rmw_connext/blob/master/rmw_connext_shared_cpp/src/shared_functions.cpp#L582-L601) are never reset to false.

As a result, if you've ever done something to trigger a fixed guard condition (e.g., add a node or subscriber), then `rmw_wait()` on Connext should return immediately, every time you call it, without waiting. That behavior might explain many of our test failures.

The proposed fix is to add a block to the end of the wait call to reset the fixed guard conditions.

gerkey:
Should fix #125.

CI jobs (connext-only) are in the queue.

### DIFF
--- a/rmw_connext_shared_cpp/include/rmw_connext_shared_cpp/shared_functions.hpp
+++ b/rmw_connext_shared_cpp/include/rmw_connext_shared_cpp/shared_functions.hpp
@@ -224,6 +224,12 @@ wait(const char * implementation_identifier,
           DDS::GuardCondition * fixed_guard_cond = static_cast<DDSGuardCondition *>(
             waitset->fixed_guard_conditions->guard_conditions[j]);
           if (fixed_guard_cond == (*attached_conditions)[i]) {
+            // Reset the fixed guard conditions to avoid being woken up
+            // immediately next time.
+            retcode = fixed_guard_cond->set_trigger_value(DDS_BOOLEAN_FALSE);
+            if (retcode != DDS_RETCODE_OK) {
+              fprintf(stderr, "failed to set trigger value\n");
+            }
             fixed = true;
             break;
           }


rmw_connext_shared_cpp/include/rmw_connext_shared_cpp/shared_functions.hpp
```diff
@@ -229,6 +229,13 @@ wait(const char * implementation_identifier,
           }
         }
         if (fixed) {
+          // Reset the fixed guard conditions to avoid being woken up
+          // immediately next time.
+          DDSGuardCondition * guard = static_cast<DDSGuardCondition *>((*attached_conditions)[i]);
+          retcode = guard->set_trigger_value(DDS_BOOLEAN_FALSE);
+          if (retcode != DDS_RETCODE_OK) {
+            RMW_SET_ERROR_MSG("failed to set trigger value");
```
wjwwood: You need to return after this I think. Otherwise the error would be ignored.

gerkey: I thought about that, but this is in the destructor for a stack variable that's being torn down on the way out of the function that declares it. I don't have a way to return an error code, and returning early would potentially avoid other useful cleanup that's happening in the destructor.

gerkey:
* http://ci.ros2.org/job/ci_linux/944/
* http://ci.ros2.org/job/ci_osx/790/
* http://ci.ros2.org/job/ci_windows/1035/

rmw_connext_shared_cpp/include/rmw_connext_shared_cpp/shared_functions.hpp
wjwwood: I see, then maybe it should go out to stderr directly? Or maybe we can just ignore it, since our error handling will print out to stderr if the message is overwritten without being read first, e.g. (problem I'm currently debugging):

```
[rmw|error_handling.c:39] error string being overwritten: wait set is invalid, at /Users/william/ros2_ws/src/ros2/rcl/rcl/src/rcl/wait.c:382
[rmw|error_handling.c:39] error string being overwritten: failed to create topic, at /Users/william/ros2_ws/src/ros2/rmw_opensplice/rmw_opensplice_cpp/src/rmw_publisher.cpp:137
/Users/william/ros2_ws/src/ros2/rcl/rcl/test/rcl/test_subscription.cpp:169: Failure
Value of: ret
  Actual: 1
Expected: 0
failed to create topic, at /Users/william/ros2_ws/src/ros2/rmw_opensplice/rmw_opensplice_cpp/src/rmw_publisher.cpp:137, at /Users/william/ros2_ws/src/ros2/rcl/rcl/src/rcl/publisher.c:81
```

You can see two messages were ignored for some reason.

dirk-thomas: And if the function can't signal the error nobody is going to check the error message. +1 to print to stderr right there.

gerkey: Ah, I had assumed that `RMW_SET_ERROR_MSG` would produce console output. What are we doing instead? `fprintf(stderr,...)`, `std::cerr`, or do we have some wrapper?

wjwwood: It places the error string in a thread specific storage buffer and the calling code can fetch it when it receives the non-ok return code. So you set the error, return an error code, the calling code notices, and finally fetches the error. It's a common pattern for passing more information along in C error handling.

I'd recommend using `fprintf(stderr, "%s\n", ...);`.

gerkey: OK, I made that change, and while I was at it, moved the reset logic to happen slightly earlier, to avoid having to recast the guard condition pointer later (no change in behavior). I accidentally squashed that last commit before pushing; sorry to do that while it's under review.

gerkey:
CI on Linux and OSX are green. Windows has 6 fewer tests than last time (I can't promise that this change had that effect, but it's possible).

wjwwood:
lgtm

PR #131, opened 2016-02-12
null not triggered conditions on timeout

dirk-thomas:
Otherwise `rclcpp` considers all conditions "active".

Based on what @gerkey discovered while looking into ros2/rclcpp#192.

http://ci.ros2.org/job/ci_linux/961/
http://ci.ros2.org/job/ci_osx/801/
Waiting for http://ci.ros2.org/job/ci_windows/1052/


gerkey:
+1

tfoote:
+1 it looks like the CI passed except for 4 on windows which look to be preexisting

PR #132, opened 2016-02-12
Fix export

esteve:
Only export library if installing.

### DIFF
--- a/rosidl_typesupport_connext_c/cmake/rosidl_typesupport_connext_c_generate_interfaces.cmake
+++ b/rosidl_typesupport_connext_c/cmake/rosidl_typesupport_connext_c_generate_interfaces.cmake
@@ -245,14 +245,19 @@ if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
     )
   endif()
 
+  if(
+    NOT "${_generated_msg_files}${_generated_external_msg_files} " STREQUAL " " OR
+    NOT "${_generated_srv_files}${_generated_external_srv_files} " STREQUAL " "
+  )
+    ament_export_include_directories(include)
+  endif()
+
   install(
     TARGETS ${rosidl_generate_interfaces_TARGET}${_target_suffix}
     ARCHIVE DESTINATION lib
     LIBRARY DESTINATION lib
     RUNTIME DESTINATION bin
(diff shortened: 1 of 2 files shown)


esteve:
CI:

http://ci.ros2.org/job/ci_linux/965/
http://ci.ros2.org/job/ci_osx/804/
http://ci.ros2.org/job/ci_windows/1057/

esteve:
Retriggered Windows because of a filesystem error:

http://ci.ros2.org/job/ci_windows/1058/

dirk-thomas:
The include directories should also not been exported when none are installed.

esteve:
@dirk-thomas done: https://github.com/ros2/rmw_connext/commit/ada700f5363d2b97c2dd9efa4469758cccefeed0

wjwwood:
Same comment as https://github.com/ros2/rmw_opensplice/pull/113, I think we could consider these for merging now. +1

rosidl_typesupport_connext_c/cmake/rosidl_typesupport_connext_c_generate_interfaces.cmake
```diff
@@ -251,8 +251,14 @@ if(NOT rosidl_generate_interfaces_SKIP_INSTALL)
...
-ament_export_libraries(${rosidl_generate_interfaces_TARGET}${_target_suffix} ${Connext_LIBRARIES})
+  ament_export_libraries(${rosidl_generate_interfaces_TARGET}${_target_suffix} ${Connext_LIBRARIES})
 
-ament_export_include_directories(include)
+  if(
+    NOT "${_generated_msg_files}${_generated_external_msg_files} " STREQUAL " " OR
+    NOT "${_generated_srv_files}${_generated_external_srv_files} " STREQUAL " "
+  )
+    ament_export_include_directories(include)
+  endif()
```
dirk-thomas: If this block would be be moved directly under the corresponding install calls it would be easier to understand the logic.

Same below.

esteve: I moved this block right after the install call for the generated msg and srv files.

PR #133, opened 2016-02-12
Select between introspection_c and introspection_cpp

jacquelinekay:
These changes are necessary to use rcl with Connext Dynamic.

### DIFF
--- a/rmw_connext_dynamic_cpp/CMakeLists.txt
+++ b/rmw_connext_dynamic_cpp/CMakeLists.txt
@@ -23,10 +23,11 @@ if(NOT Connext_FOUND OR CONNEXT_DYNAMIC_DISABLE)
   return()
 endif()
 
+find_package(rosidl_typesupport_introspection_c)
 find_package(rosidl_typesupport_introspection_cpp)
-if(NOT rosidl_typesupport_introspection_cpp_FOUND)
+if(NOT rosidl_typesupport_introspection_c_FOUND AND NOT rosidl_typesupport_introspection_cpp_FOUND)
   ament_package()
-  message(WARNING "Could not find ROSIDL TypeSupport for Connext (rosidl_typesupport_introspection_cpp) - skipping '${PROJECT_NAME}'")
+  message(WARNING "Could not find ROSIDL TypeSupport for Connext (rosidl_typesupport_introspection_c|cpp) - skipping '${PROJECT_NAME}'")
   return()
 endif()
 
@@ -45,6 +46,7 @@ register_rmw_implementation("rosidl_typesupport_introspection_c;rosidl_typesuppo
 
 add_library(rmw_connext_dynamic_cpp SHARED src/functions.cpp)
 ament_target_dependencies(rmw_connext_dynamic_cpp
+  "rosidl_typesupport_introspection_c"
   "rosidl_typesupport_introspection_cpp"
   "rmw"
   "rmw_connext_shared_cpp"
--- a/rmw_connext_dynamic_cpp/src/functions.cpp
+++ b/rmw_connext_dynamic_cpp/src/functions.cpp
@@ -48,6 +48,9 @@
 // is in the include/rosidl_typesupport_connext_cpp/impl folder.
 #include "rosidl_generator_cpp/message_type_support.hpp"
 
+#include "rosidl_generator_c/string.h"
+#include "rosidl_generator_c/string_functions.h"
+
 #include "rmw/impl/cpp/macros.hpp"
 
 #include "rosidl_typesupport_introspection_cpp/field_types.hpp"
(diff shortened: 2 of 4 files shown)


rmw_connext_dynamic_cpp/CMakeLists.txt
```diff
@@ -23,10 +23,11 @@ if(NOT Connext_FOUND OR CONNEXT_DYNAMIC_DISABLE)
   return()
 endif()
 
+find_package(rosidl_typesupport_introspection_c)
 find_package(rosidl_typesupport_introspection_cpp)
-if(NOT rosidl_typesupport_introspection_cpp_FOUND)
+if(NOT rosidl_typesupport_introspection_cpp_FOUND AND NOT rosidl_typesupport_introspection_c)
```
dirk-thomas: Lacks `_FOUND` suffix?

wjwwood:
screen shot 2016-02-12 at 2 41 40 pm

lol

jacquelinekay:
That's how we roll.

jacquelinekay:
I'm starting CI for this to see that it doesn't break anything.

http://ci.ros2.org/job/ci_linux/978
http://ci.ros2.org/job/ci_osx/813
http://ci.ros2.org/job/ci_windows/1064 (I think)

I haven't tested this with C yet though. I'll make a differently named branch to enable building rcl for Connext Dynamic.

jacquelinekay:
Alright, I fixed some stuff with strings and started a branch in rcl that enables the dynamic test again.

http://ci.ros2.org/job/ci_linux/982

jacquelinekay:
I made some style fixes, relaunched with only Connext Dynamic enabled (hope that works).

http://ci.ros2.org/job/ci_linux/983/

jacquelinekay:
http://ci.ros2.org/job/ci_linux/994
http://ci.ros2.org/job/ci_osx/823/
http://ci.ros2.org/job/ci_windows/1074

jacquelinekay:
Fixed a warning on OSX and Windows.

http://ci.ros2.org/job/ci_osx/827/

I think the failed tests on Windows also appear on the nightly build. I think we should fix those failures before merging this, but we should start review on this now.

I separated the new macros into a macros.hpp file so that the diff isn't impossible to view in the browser.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -56,17 +59,45 @@
...
-  const std::string & sep)
+  const void * untyped_members,
+  const std::string & sep,
+  const char * typesupport)
 {
+  std::string package_name, message_name;
+  if (using_introspection_c_typesupport(typesupport)) {
+    CREATE_TYPENAME_PREFIX(INTROSPECTION_C_TYPE)
+  } else if (using_introspection_cpp_typesupport(typesupport)) {
+    CREATE_TYPENAME_PREFIX(INTROSPECTION_CPP_TYPE)
```
dirk-thomas: Since both macros have the code to access the fields of `members` in common I would suggest to let the macro only define the `members` variable and continue to handle accessing the fields `package_name_` and `message_name_` outside.

dirk-thomas: Then the macro could also be renamed to something like `GET_TYPED_MEMBERS`.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -188,165 +225,17 @@ DDS_TypeCode * create_type_code(
...
-      RMW_SET_ERROR_MSG("failed to get primitive typecode");
-      goto fail;
-    }
-    type_code->add_member("_dummy", DDS_TYPECODE_MEMBER_ID_INVALID,
-      member_type_code,
-      DDS_TYPECODE_NONKEY_REQUIRED_MEMBER, ex);
-    if (ex != DDS_NO_EXCEPTION_CODE) {
-      RMW_SET_ERROR_MSG("failed to add member");
-      goto fail;
-    }
```
dirk-thomas: Since moving such big blocks of code makes it very difficult to review code changes I would suggest to have a separate commit (which doesn't get squashed) which only moves the code from the one location to the new location (without changing it). Then this specific commit can be ignored after a short glance.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -878,7 +762,59 @@ rmw_destroy_publisher(rmw_node_t * node, rmw_publisher_t * publisher)
...
+      DDS_ReturnCode_t status = dynamic_data->METHOD_NAME( \
+        NULL, \
+        static_cast<DDS_DynamicDataMemberId>(i + 1), \
+        value->data); \
+      if (status != DDS_RETCODE_OK) { \
+        RMW_SET_ERROR_MSG("failed to set value using " #METHOD_NAME); \
+        return false; \
+      } \
+    } \
+  }
```
dirk-thomas: Since this is a block of pretty complicated code how is this different to `SET_CPP_STRING_VALUE`?

jacquelinekay:
After some thought I created a branch with a different approach that is more concise than this one (although I'm still not sure if it's the best):

https://github.com/ros2/rmw_connext/compare/introspection_template

Unfortunately I'm running into a compile error that I can't quite figure out, due to 7e8232e. Looks like feature isn't going to make it into this release, if the pull request in the current state is stylistically unacceptable.

wjwwood:
@jacquelinekay thanks for trying to get it in. If you need someone to review it, then I can lend a hand after the alpha. We can try to get it merged asap after the release.

jacquelinekay:
alright, I took a pretty different approach that is more concise and I reset this branch to reflect it. (the old approach is still around, in https://github.com/ros2/rmw_connext/compare/introspection_cpp_old)

http://ci.ros2.org/job/ci_linux/1020/
http://ci.ros2.org/job/ci_osx/846/
http://ci.ros2.org/job/ci_windows/1087/

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -56,19 +59,112 @@
...
+
 #include "rmw_connext_shared_cpp/shared_functions.hpp"
 #include "rmw_connext_shared_cpp/types.hpp"
 
+#include "macros.hpp"
+
+
+bool using_introspection_c_typesupport(const char * typesupport_identifier)
+{
+  return strcmp(typesupport_identifier, rosidl_typesupport_introspection_c__identifier) == 0;
```
dirk-thomas: The should not use string comparison since it is a pretty expensive operation. See the previously used macro: https://github.com/ros2/rmw/blob/55254b66ddb0f7c5ff8f05fb8c0c60378fb3b2da/rmw/include/rmw/impl/cpp/macros.hpp#L70

Same below.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -56,19 +59,112 @@
...
 inline std::string
 _create_type_name(
-  const rosidl_typesupport_introspection_cpp::MessageMembers * members,
+  const void * untyped_members,
   const std::string & sep)
 {
+  auto members = static_cast<const MembersType *>(untyped_members);
+  if (!members) {
+    RMW_SET_ERROR_MSG("members handle is null");
+    return NULL;
```
dirk-thomas: Doesn't fit the return type.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -56,19 +59,112 @@
...
+  const char * typesupport)
+{
+  if (using_introspection_c_typesupport(typesupport)) {
+    return _create_type_name<rosidl_typesupport_introspection_c__MessageMembers>(untyped_members,
+             sep);
+  } else if (using_introspection_cpp_typesupport(typesupport)) {
+    return _create_type_name<rosidl_typesupport_introspection_cpp::MessageMembers>(untyped_members,
+             sep);
+  }
+  RMW_SET_ERROR_MSG("service members handle is null");
```
dirk-thomas: The error message doesn't match the error case.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -439,14 +473,20 @@ rmw_create_publisher(
   DDSDynamicDataWriter * dynamic_writer = nullptr;
   DDS_DynamicData * dynamic_data = nullptr;
   CustomPublisherInfo * custom_publisher_info = nullptr;
+
+  std::string type_name = _create_type_name(type_support->data, "msg",
+      type_support->typesupport_identifier);
+
   // Start initializing elements.
+
```
dirk-thomas: Please avoid unrelated white space changes.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -1210,6 +1179,8 @@ rmw_create_subscription(
   DDS_DataReaderQos datareader_qos;
   DDS_DynamicData * dynamic_data = nullptr;
   CustomSubscriberInfo * custom_subscriber_info = nullptr;
+
+
```
dirk-thomas: Please avoid unrelated whitespace changes.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -2165,6 +2033,40 @@ rmw_create_client(
...
+  std::string request_type_name;
+  std::string response_type_name;
+  const void * untyped_request_members;
+  const void * untyped_response_members;
+
+  DDS_DomainParticipantQos participant_qos;
+  DDS_ReturnCode_t status = participant->get_qos(participant_qos);
+  if (status != DDS_RETCODE_OK) {
+    RMW_SET_ERROR_MSG("failed to get participant qos");
+    return NULL;
```
dirk-thomas: Since this part return early it can stay above all the other variable declarations. No need to move it down.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -2455,7 +2347,8 @@ rmw_send_request(
   DDS::WriteParams_t writeParams;
   connext::WriteSampleRef<DDS::DynamicData> request(*sample, writeParams);
 
-  bool published = _publish(sample, ros_request, client_info->request_members_);
+  bool published = _publish(sample, ros_request, client_info->untyped_request_members_,
```
dirk-thomas: Please wrap after open parenthesis if any of the arguments is wrapped later.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -2165,6 +2033,40 @@ rmw_create_client(
...
+      type_support->typesupport_identifier);
+  response_type_name = _create_type_name(untyped_response_members, "srv",
+      type_support->typesupport_identifier);
+
+  request_type_code = _create_type_code(request_type_name, untyped_request_members,
+      type_support->typesupport_identifier);
+  if (!request_type_code) {
+    // error string was set within the function
+    goto fail;
+  }
```
dirk-thomas: While this generic error condition works the error message will be less specific (just mentioning that some members are null without specifying which one).

Same below.

dirk-thomas:
To simplify the code further I think there should be only one list of field types (in C) and C++ should use the same constant.

The C++ part could still define "nicer" constants in a namespace (based on the C constants) but the rmw impl. would only need to deal with the C constants.

dirk-thomas:
Overall this looks much nicer then the previous branch.

rmw_connext_dynamic_cpp/src/functions.cpp
jacquelinekay: everything here is a const char *, so using the equality operator should be okay.

jacquelinekay:
I agree that the C/C++ introspection APIs should share some types; I see no reason why they need separate definitions--that was the most frustrating thing about making this change...

I suppose the current scheme is better if we wanted to avoid a dependency of rosidl_typesupport_introspection_cpp on rosidl_typesupport_introspection_c, but I think most dynamic RMW implementations will want to support both C and C++ introspection and will therefore depend on both packages. So introducing a dependency between the two introspection packages is not a big deal.

jacquelinekay:
Oh, I understand now, changes to the introspection packages aren't necessarily needed, I believe the field type enums from either package can be used in the switch statements that I currently have macros for.

jacquelinekay:
I realized why sharing the enum definitions would be nice to have. ros2/rosidl#94

rmw_connext_dynamic_cpp/src/functions.cpp
jacquelinekay: I actually think this return type is OK. The return type of this function is a pointer and on an error it returns a null pointer.

dirk-thomas: I was reading its supposed to return a string. No idea why. Looks good as it is. Sorry for the noise.

jacquelinekay:
CI for the new approach, with only connext dynamic enabled.

http://ci.ros2.org/job/ci_linux/1029
http://ci.ros2.org/job/ci_osx/848/
http://ci.ros2.org/job/ci_windows/1088/

esteve:
+1

jacquelinekay:
Since @esteve gave his +1 I'll merge these PRs today unless there are any objections.

esteve:
:fireworks: :tada: :balloon:

Issue #63, opened 2015-07-29
[rmw_connext_dynamic] fix cleanup functions


dirk-thomas:
In #45 various clean up functions have been introduced. In various tests (e.g. in `test_rclcpp`) these functions print error messages:

```
Error in destruction of rmw publisher handle: failed to delete contained entities for publisher, at ...
Error in destruction of rmw subscription handle: failed to delete subscriber, at ...
```

wjwwood:
I guess I didn't notice the output of the tests since they were passing. I would expect that to cause a failure (return code != 0 or exception) since those messages occur the `rmw_*_destroy` functions. Probably they're being checked in a C++ destructor and being printed to the screen rather than throwing to avoid throwing in a destructor.

jacquelinekay:
I'm running into an issue where a test case that initializes services/clients in a gtest with prior cases with pub/sub does not send any messages. This may be related, so I'm going to investigate cleanup issues in rmw and see if it helps.

dirk-thomas:
Furthermore the Connext typesupport  does currently not provide any destroy functions for servers / clients.

clalancette:
Closing, since with Foxy now being [End-of-Life](https://discourse.ros.org/t/new-packages-for-foxy-fitzroy-2023-06-20/32039), this repository is no longer used.

PR #134, opened 2016-02-13
fix dynamic client

dirk-thomas:
While this disables some cleanup it makes ros2/rclcpp#192 pass. Since rmw_connext (static) does not implement any server / client destroy functionality I just added a todo for this case. I also added a comment for this in the existing ticket: ros2/rmw_connext#63

* http://ci.ros2.org/job/ci_linux/981/
* http://ci.ros2.org/job/ci_osx/815/

### DIFF
--- a/rmw_connext_dynamic_cpp/src/functions.cpp
+++ b/rmw_connext_dynamic_cpp/src/functions.cpp
@@ -2192,7 +2192,7 @@ rmw_create_client(
   buf = nullptr;  // Only free the casted pointer; don't need the buf pointer anymore.
 
   response_type_code = create_type_code(response_type_name, response_members, participant_qos);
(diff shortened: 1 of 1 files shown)


tfoote:
This looks a bit hacky but if it gets the unit tests passing and we've ticketed the fixes I guess it's ok.

gerkey:
Were you able to determine whether this change leaks memory when service clients are repeatedly created and destroyed?

Also, what's the reason for taking out the qos arg?

dirk-thomas:
I am pretty sure that it will not free the memory. I can't tell of the participant would clean it up at the end though.

The `create_type_code` functions simply doesn't use the argument. So no need to pass it.

gerkey:
Good enough, at least for now. +1

PR #135, opened 2016-02-16
Enable service typesupport for C

jacquelinekay:
I believe this will fix an issue @dirk-thomas found with compiling packages with only services.

Because the template file for C services typesupport is currently empty, services won't actually work yet.


jacquelinekay:
Confirmed locally by removing the AMENT_IGNORE file in std_srvs, a package in common_interfaces which only has services.

dirk-thomas:
+1 with another package containing only service files but no messages.

jacquelinekay:
ros2/common_interfaces#10

Do you think re-enabling std_srvs is the right approach, or should we make a smaller new package somewhere?

dirk-thomas:
I don't think reenabling `std_srvs` is necessary neither create a new package for it. Simply because we disabled for faster build times and I think this is still an important goal. So I would go for no coverage for that case for the near future (until we reenable more packages again).

jacquelinekay:
http://ci.ros2.org/job/ci_linux/1000/
http://ci.ros2.org/job/ci_osx/831/
http://ci.ros2.org/job/ci_windows/1081 (pending)

I kept the branch that enables std_srvs for this build, but I will not merge it (and I will delete the branch after merging).

PR #137, opened 2016-02-24
Add topic name

jacquelinekay:
Connects to ros2/rcl#29

Without this PR, the test in the above PR will fail with Connext and Connext dynamic.


wjwwood:
lgtm

PR #140, opened 2016-03-01
Services typesupport cpp

jacquelinekay:
Improve Connext client/service type support by not type-erasing the request header, adding an allocate function to create_requester/create_replier, and implementing destroy_requester and destroy_replier functions.

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -878,7 +878,7 @@ rmw_create_client(
 
   requester = callbacks->create_requester(
     participant, service_name, &datareader_qos, &datawriter_qos,
-    reinterpret_cast<void **>(&response_datareader));
+    reinterpret_cast<void **>(&response_datareader), &rmw_allocate);
   if (!requester) {
     RMW_SET_ERROR_MSG("failed to create requester");
     goto fail;
@@ -964,6 +964,12 @@ rmw_destroy_client(rmw_client_t * client)
     } else if (client_info->read_condition_) {
       RMW_SET_ERROR_MSG("cannot delete readcondition because the datareader is null");
       result = RMW_RET_ERROR;
+    }
+    const service_type_support_callbacks_t * callbacks = client_info->callbacks_;
+    if (callbacks) {
+      if (client_info->requester_) {
+        callbacks->destroy_requester(client_info->requester_, &rmw_free);
+      }
     }
 
     RMW_TRY_DESTRUCTOR(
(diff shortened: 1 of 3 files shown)


jacquelinekay:
http://ci.ros2.org/job/ci_linux/1044/
http://ci.ros2.org/job/ci_osx/861/
http://ci.ros2.org/job/ci_windows/1105/

wjwwood:
The code changes lgtm and the CI looks good for Linux and OS X. For Windows there appear to be some new warnings, e.g.:

http://ci.ros2.org/job/ci_windows/1105/warnings34Result/new/package.1133270863/file.-93741396/

Looks like Windows thinks the catching of the explicit destructor call is unreachable? Not sure why that's a warning.

Other than the new warnings on Windows, +1.

jacquelinekay:
I'll address the Windows warning and retrigger CI.

jacquelinekay:
The try/catch pattern for explicitly calling the destructor is copied from rosidl_typesupport_opensplice_cpp:

https://github.com/ros2/rmw_opensplice/blob/master/rosidl_typesupport_opensplice_cpp/resource/srv__type_support.cpp.template#L528-L536

If the destructor for the Connext Replier were annotated as `noexcept`, then the catch block would be considered unreachable code.

But I couldn't find anything in the Connext documentation to indicate that the destructor of Requester or Replier is `noexcept`.

dirk-thomas:
I destructor in C++ can't throw an exception. See http://c2.com/cgi/wiki?BewareOfExceptionsInTheDestructor

jacquelinekay:
I believe destructors in C++ can throw exceptions, but the page you linked to is giving good reasons why we shouldn't write destructors that can throw exceptions.

See bottom of this page: http://en.cppreference.com/w/cpp/language/destructor

"As any other function, a destructor may terminate by throwing an exception (this usually requires it to be explicitly declared noexcept(false)) (since C++11), however if this destructor happens to be called during stack unwinding, std::terminate is called instead."

jacquelinekay:
I think it's safe to remove the try/catch block around the lines that are giving warnings in Windows.

I still find it strange because those lines are vendor code (Connext Requester/Replier API), and the original try/catch block in rosidl_typesupport_opensplice_cpp is for code that we control which is explicitly marked noexcept... yet we don't see any warnings from Windows about the opensplice code.

http://ci.ros2.org/job/ci_windows/1111/

jacquelinekay:
The new warnings are fixed (back to 60 warnings) so I will squash and merge this.

PR #141, opened 2016-03-01
C services

jacquelinekay:
Connects to ros2/rcl#27
based on #140

Yet another C services pull request.

Provide a generated library that creates the expected C structures for service-related callbacks and calls directly to the C++ implementation.

Again it's a bit strange that C type support depends so much on C++ type support. Perhaps we should reverse the dependency so that C++ depends on C. Regardless, I think that the principal of it is fairly sound to reduce code duplication.

### DIFF
--- a/rosidl_typesupport_connext_cpp/resource/srv__type_support.hpp.template
+++ b/rosidl_typesupport_connext_cpp/resource/srv__type_support.hpp.template
@@ -1,0 +1,92 @@
+// generated from rosidl_typesupport_connext_cpp/resource/srv__type_support.hpp.template
+// generated code does not contain a copyright notice
+
+@#######################################################################
+@# EmPy template for generating <srv>__type_support.hpp files
+@#
+@# Context:
+@#  - spec (rosidl_parser.MessageSpecification)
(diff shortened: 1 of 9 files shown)


jacquelinekay:
this doesn't totally work yet. I need to change the generate_interfaces.cmake to properly depend on the interfaces generated by rosidl_generator_cpp and rosidl_typesupport_connext_cpp...

jacquelinekay:
should be working now

rosidl_typesupport_connext_cpp/resource/srv__type_support.hpp.template
```diff
@@ -0,0 +1,91 @@
+// generated from rosidl_typesupport_connext_cpp/resource/srv__type_support.hpp.template
```
dirk-thomas: Needs "no copyright" line.

wjwwood:
Just finished reviewing the code and I don't have any comments to add. lgtm

jacquelinekay:
I'm going to relaunch CI just in case then squash and merge all c_services branches if clean.

PR #143, opened 2016-03-03
suppress warning for generated Connext code

dirk-thomas:
This will fix the following compiler warnings: http://ci.ros2.org/job/ci_linux/1051/warnings17Result/category.-2074711528/

### DIFF
--- a/rosidl_typesupport_connext_c/cmake/rosidl_typesupport_connext_c_generate_interfaces.cmake
+++ b/rosidl_typesupport_connext_c/cmake/rosidl_typesupport_connext_c_generate_interfaces.cmake
@@ -77,6 +77,7 @@ if(NOT WIN32)
     )
   elseif("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
     set(_connext_compile_flags
+      "-Wno-strict-aliasing"
       "-Wno-unused-but-set-variable"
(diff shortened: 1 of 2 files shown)


dirk-thomas:
The second commit fixes the following compiler warning: http://ci.ros2.org/job/ci_linux/1050/warnings17Result/

wjwwood:
lgtm. Should we be doing CI, or are we just going to merge each proposed fix and do a CI at the end of the sprint?

dirk-thomas:
I ran it locally before and after which fixed the compiler warning. So I would propose to merge and run a CI afterwards to also cover other changes merged in the meantime and keep the load on the farm low.

dirk-thomas:
Two generators, same in both. Committed f645b9bd1d3cdb342ba5d51df9c46830eddff19b

jacquelinekay:
why do the strict-aliasing warnings still appear in the Release build?

http://ci.ros2.org/view/nightly/job/nightly_linux_release/1/warnings17Result/new/package.-530164750/

dirk-thomas:
This PR only modified `rosidl_typesupport_connext_c`. The same fix is needed for `rosidl_typesupport_connext_cpp`.

Based on my previous comment (`Two generators, same in both`) I am pretty sure it was changed for both at some point. Anyway this is the follow up PR: #146

PR #145, opened 2016-03-04
modified generated Connext code to suppress unused global variable warning

dirk-thomas:
Before: http://ci.ros2.org/job/ci_linux/1065/
After: http://ci.ros2.org/job/ci_linux/1066/


tfoote:
lgtm

PR #146, opened 2016-03-09
fix stric-aliasing warning

dirk-thomas:
Before: http://ci.ros2.org/view/nightly/job/nightly_linux_release/6/warnings17Result/
After: http://ci.ros2.org/job/ci_linux/1091/


jacquelinekay:
+1. btw, I believe those additional test failures will be fixed by ros2/examples#94 and ros2/rclpy#6, which I merged after you launched those jobs.

PR #147, opened 2016-03-17
Remove fixed guard conditions

jacquelinekay:
Connects to ros2/rcl#33


wjwwood:
This is subject to the discussion in https://github.com/eProsima/ROS-RMW-Fast-RTPS-cpp/pull/19#discussion_r57213635, but otherwise it lgtm.

PR #148, opened 2016-03-31
Change semantics of how rmw arrays are interpreted

jacquelinekay:
I discussed with @wjwwood online after running into some issues integrating rcl and rclcpp and we decided that the rmw implementations need to interpret the arrays of subscriptions, guard conditions, and clients are interpreted, to match how rcl allocates and fills the rmw arrays that it passes to `rmw_wait`. Instead of treating the `..._count` field as the number of valid entries in the array, `rmw_wait` now treats the `count` as the capacity of the array, and stops iterating over the entries of the array when a null entry is encountered.

### DIFF
--- a/rmw_connext_shared_cpp/include/rmw_connext_shared_cpp/shared_functions.hpp
+++ b/rmw_connext_shared_cpp/include/rmw_connext_shared_cpp/shared_functions.hpp
@@ -279,11 +279,15 @@ wait(const char * implementation_identifier,
   }
 
   // add a condition for each subscriber
-  for (size_t i = 0; i < subscriptions->subscriber_count; ++i) {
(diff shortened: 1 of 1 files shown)


dirk-thomas:
Should we rely on the elements always to be ordered? The loop could just skip null element rather then to break.

wjwwood:
> Should we rely on the elements always to be ordered? The loop could just skip null element rather then to break.

I was going to suggest the same thing. The downside is if the capacity of the array is large and the actual elements in it are very small (e.g. capacity is 1000 but there is only one element in the 0 index), then we'll iterate unnecessarily over the entire 1000.

A more complicated solution would be to have capacity, size, and allow skipping over null elements up to the index `size - 1`.

jacquelinekay:
After a discussion offline we decided to keep the semantics in rmw as they are currently and change rcl. I will close this and make a PR to add documentation to the rmw header file.

PR #149, opened 2016-04-06
suppress deprecated declaration warning for Connext code

dirk-thomas:
Suppress deprecated declaration warnings within Connext with newer G++ compiler and use old ABI in order to be compatible with Connext binaries.

* http://ci.ros2.org/job/ci_linux/1161/
* http://ci.ros2.org/job/ci_osx/939/
* http://ci.ros2.org/job/ci_windows/1217/

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -25,6 +25,7 @@
 
 #ifndef _WIN32
 # pragma GCC diagnostic push
+# pragma GCC diagnostic ignored "-Wdeprecated-declarations"
 # pragma GCC diagnostic ignored "-Wunused-parameter"
 # ifdef __clang__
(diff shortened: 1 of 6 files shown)


jacquelinekay:
Is this for upgrading to the default gcc version in Xenial?

As a side note, if you have more things in mind that need to be fixed when we upgrade to Xenial, maybe we could make a ticket to aggregate them. There are a few things in rclcpp that could be cleaned up with the better C++11 support in the next g++ version.

dirk-thomas:
Yes, this is with the default compiler on Xenial.

I don't know (yet) what else might come up. So I Ilene have anything to fill an aggregated ticket.

I don't think that the Xenial compiler is relevant for C++11 support. The "lowest" bar is still Visual Studio as far as I can see.

jacquelinekay:
+1

PR #151, opened 2016-04-15
Make identifier consistent with package name

jacquelinekay:
Make the hardcoded string identifier consistent with the package name. This is to avoid a hardcoded map in any situation where we want to match the identifier to the rmw implementation package name, e.g. https://github.com/ros2/rosidl/blob/master/rosidl_generator_py/rosidl_generator_py/import_type_support_impl.py#L17-L20

Perhaps the most proper thing to do would be to register an ament index entry to be used as the identifier (I discussed this with @wjwwood a while ago). This is a simpler fix for now (and avoids introducing a dependency on the ament index and introducing file IO).

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -64,7 +64,7 @@ _create_type_name(
 
 extern "C"
 {
-const char * rti_connext_identifier = "connext_static";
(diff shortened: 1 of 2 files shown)


tfoote:
+1

declaring it in the ament index might make sense too for external tool visibility, but I don't think it should have to look up it's own name in the ament index.

PR #154, opened 2016-04-19
use ensure_length instead of length for memory allocation

mikaelarguedas:
Connects to https://github.com/ros2/rosidl/issues/114

### DIFF
--- a/rosidl_typesupport_connext_c/resource/msg__type_support_c.cpp.template
+++ b/rosidl_typesupport_connext_c/resource/msg__type_support_c.cpp.template
@@ -182,7 +182,7 @@ convert_ros_to_dds(const void * untyped_ros_message, void * untyped_dds_message)
       return false;
     }
     DDS_Long length = static_cast<DDS_Long>(size);
-    dds_message->@(field.name)_.length(length);
+    dds_message->@(field.name)_.ensure_length(length, length);
 @[end if]@
     for (DDS_Long i = 0; i < static_cast<DDS_Long>(size); ++i) {
 @[if field.type.array_size]@
--- a/rosidl_typesupport_connext_cpp/resource/msg__type_support.cpp.template
+++ b/rosidl_typesupport_connext_cpp/resource/msg__type_support.cpp.template
@@ -113,7 +113,7 @@ convert_ros_message_to_dds(
       throw std::runtime_error("failed to set maximum of sequence");
     }
 @[end if]@
-    if (!dds_message.@(field.name)_.length(length)) {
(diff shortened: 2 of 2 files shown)


rosidl_typesupport_connext_c/resource/msg__type_support_c.cpp.template
```diff
@@ -182,7 +182,7 @@ convert_ros_to_dds(const void * untyped_ros_message, void * untyped_dds_message)
       return false;
     }
     DDS_Long length = static_cast<DDS_Long>(size);
-    dds_message->@(field.name)_.length(length);
+    dds_message->@(field.name)_.ensure_length(length, 2 * length);
```
dirk-thomas: Why allocate twice as much memory then needed?

wjwwood: That's the max, it only allocates if there is not enough memory of length. I suppose, instead, it should be something like `current_capacity * 2`.

wjwwood: If the `dds_message` is temporary, then as @dirk-thomas suggested it should be `length, length`. @mikaelarguedas can you check to make sure this message is temporary and update if it is?

dirk-thomas: I am not sure if the `dds_message` instance exists longer then for a single serialization but if not I think `(length, length)` would be better to not allocate more memory then needed.

mikaelarguedas: Yes that's my feeling too, the message should be temporary, didn't have time to look more into it but will give an update soon

mikaelarguedas: Message is indeed temporary. Setting the allocated memory to `length`

jacquelinekay:
There's similar calls in `rosidl_typesupport_connext_cpp`, does it need to be changed there as well?

https://github.com/ros2/rmw_connext/blob/master/rosidl_typesupport_connext_cpp/resource/msg__type_support.cpp.template#L116

mikaelarguedas:
Yeah it would make sense in the case that `is_upper_bound` is set and that the current `maximum` is less than the upper bound. Then I don't know if this scenario can actually happen.
In the end I would say it's safer to change it anyway.

dirk-thomas:
+1

wjwwood:
Can this just be merged?

PR #155, opened 2016-04-19
fix memory leak

jacquelinekay:
This was driving me insane. Any example of rclcpp or rcl leaks memory from the guard condition handles (you can check with `valgrind --leak-check=full -v <executable name>`). This fixes the leaks.


wjwwood:
lgtm

PR #156, opened 2016-04-25
Add item to resource index for license output filtering

dhood:
Needed for https://github.com/ros2/launch/pull/24 and https://github.com/ros2/examples/pull/99

if approved I'll extend it to the other rmw implementations


dirk-thomas:
Since both Connext packages do the same thing the functionality should be moved (and deduplicated) to `rmw_connext_shared_cpp`.

The resource file doesn't need to be installed since it isn't used anywhere after `ament_index_register_resource` has been called. Instead of the two lines in the file the CMake code could contain them directly (not sure which is preferred).

dhood:
I understood from [this paragraph](https://github.com/ament/ament_cmake/blob/master/ament_cmake_core/doc/resource_index.md#marker-files) that it is better to install a file to `share` which contains the information (license text) instead of putting it into the resource file itself. So the installed file is being used [here](https://github.com/ros2/launch/pull/24/commits/6b3116cc6f5c05b08fddec4a8ae7b2ca9e37b1c8#diff-95fa932db2912f5a4060524ea72cd0cfR108)

That's why it is not in the `rmw_connext_shared_cpp` package, but I could put a function in it and call it from the other two packages if appropriate

dirk-thomas:
For the trivial content it would be fine to place it in the resource itelf. It is just not recommended to e.g. put the content of a plugin.xml file into the resource directly. But either or is fine I guess.

Anyway the file as well as the CMake code to register the resource can be moved (and be deduplicated) into the `rmw_connext_shared_cpp` package. It can e.g. define a CMake function which the other two Connext packages then call to register their package specific resource. But the file itself should only exist once.

dhood:
Ok I'll leave the resource as it is. 

But I see your point about the file itself being duplicated - I will move it into the shared package, and get the `rmw_connext_static_cpp` and `rmw_connext_dynamic_cpp` packages to check the `share` directory of `rmw_connext_shared_cpp` for the file. Is that how you would recommend?

I'll also make a function like [register_rmw_implementation.cmake](https://github.com/ros2/rmw/blob/master/rmw/cmake/register_rmw_implementation.cmake) in `rmw` called `register_rmw_output_filter` unless you don't think it belongs there

dirk-thomas:
What would `register_rmw_output_filter` do beside calling `ament_index_register_resource("rmw_output_filter" ...)`? If that is all I am not sure we need an explicit function for it.

dhood:
I am still approaching it from the angle where the resource file only has the path of the file with the filters in it because of the recommendation to not rely on the content of the ament index files

so the function would register the resource but also install the file in the appropriate place (so that it would be consistent for all rmw implementations)

dirk-thomas:
The current code (in roslaunch) makes the assumption about the base path of the file containing the output filter. Accessing the information from different languages will require to embed that assumption in each of them. For the simple data in this case I am not sure if the indirection is worth the effort. Also especially in C++ the file system handling is more effort and might require external dependencies like boost::filesystem.

dhood:
when accessing the filter info, isn't the only assumption that's made in [this implementation](https://github.com/ros2/launch/pull/24/commits/6b3116cc6f5c05b08fddec4a8ae7b2ca9e37b1c8#diff-95fa932db2912f5a4060524ea72cd0cfR108) that the package has a share folder? the rest of the path comes from the value inside the ament resource.

I agree that accessing the file from different languages will be more complicated than if the info were in the resource file itself. But in c++ shouldn't you just be able to access the file in the same way that [ament_indet/get_resource](https://github.com/ament/ament_index/blob/master/ament_index_cpp/src/get_resource.cpp) does (without boost::filesystem)? 

I absolutely agree that this might not be worth the indirection, I was just following the approach recommended in the `ament_index` documentation.  I think it'd be great if one day there were a function like `get_package_resource(package_name, path)` which would replace [lines 101-110 here](https://github.com/ros2/launch/pull/24/commits/6b3116cc6f5c05b08fddec4a8ae7b2ca9e37b1c8#diff-95fa932db2912f5a4060524ea72cd0cfR108) and would also exist in c++ and would generally make this indirection not take any effort. I can implement that and other `rospack`-like functions if it's appropriate, or I can just skip the indirection if you prefer.

dirk-thomas:
I would just skip the indirection since I don't see any benefit in this case. But maybe others have a different opinion? @ros2/ros2_team

wjwwood:
In my opinion, the redirection is an implementation detail of the particular resource, so I don't think loading its data needs to be supported in every language (as opposed to listing and reading files in the ament index). So if you're just accessing it in CMake or Python, and don't expect to access it from C++, then leaving the indirection is fine by me.

As @dirk-thomas said, the indirection is choice and you might choose to use it or not based on the situation. In this case, I neither see any benefit to have the indirection nor any benefit to not having it (since we're not accessing it from C++ and accessing it from Python is trivial). My slight preference would be to leave it as-is because that's the least work.

dhood:
it's no trouble to simplify it

I'd like to check something for my understanding, though: I understood from [this paragraph](https://github.com/ament/ament_cmake/blob/master/ament_cmake_core/doc/resource_index.md#integration-with-other-systems) (which it sounds like you wrote @wjwwood ) that it is better to have this (rmw output) information stored in the package's (`rmw_connext_cpp`'s) share folder so that it does not rely on `ament_index` to access it. Yes, the license output is a pretty trivial case, but I would think the concept would still apply here? That was my reasoning for the indirection. again, I'm happy to simplify it, I just wanted to understand in what cases that paragraph doesn't apply (perhaps it is simply unnecessary for ros2 core packages, for example, but is more relevant for user packages)

dhood:
I have updated this PR and propagated the changes into `launch_testing`

The CI still seems fine (the three branches connected to this PR)
http://ci.ros2.org/job/ci_windows/1285/#showFailuresLink
http://ci.ros2.org/job/ci_osx/1002/
http://ci.ros2.org/job/ci_linux/1247/

rmw_connext_dynamic_cpp/CMakeLists.txt
```diff
@@ -84,3 +84,6 @@ install(
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin
 )
+
+ament_index_get_resource(connext_shared_output_filter "rmw_output_filter" "rmw_connext_shared_cpp")
+ament_index_register_resource("rmw_output_filter" CONTENT "${connext_shared_output_filter}")
```
dirk-thomas: Please add these lines in the same location as for `rmw_connext_cpp` to keep both similar.

rmw_connext_shared_cpp/CMakeLists.txt
```diff
@@ -56,3 +56,6 @@ install(
   LIBRARY DESTINATION lib
   RUNTIME DESTINATION bin
 )
+
+ament_index_register_resource("rmw_output_filter"
+  CONTENT "RTI Data Distribution Service\nExpires on")
```
dirk-thomas: Should this information be passed through the index or just be exposed in a CMake variable / function?

dhood: I've added a cmake function now

dhood:
Could someone review this + the two connected PRs please? CI seems fine
http://ci.ros2.org/job/ci_windows/1353/ (no new failures)
http://ci.ros2.org/job/ci_linux/1339/ (no failures)
http://ci.ros2.org/job/ci_osx/1049/ (no failures)

rmw_connext_cpp/CMakeLists.txt
```diff
@@ -60,11 +67,6 @@ if(WIN32)
             "RMW_CONNEXT_SHARED_BUILDING_DLL")
 endif()
 
-register_rmw_implementation(
-  "c:rosidl_typesupport_connext_c"
-  "cpp:rosidl_typesupport_connext_cpp"
-)
-
```
wjwwood: Why did this need to move?

dhood: it doesn't need to, I just did it to make it match rmw_connext_dynamic_cpp/CMakeLists.txt

wjwwood:
They all lgtm

rmw_connext_cpp/CMakeLists.txt
wjwwood: Sounds good.

PR #159, opened 2016-05-05
ignore warning with Clang in Connext code

dirk-thomas:
Before: http://ci.ros2.org/job/ci_osx/1010/

After: http://ci.ros2.org/job/ci_osx/1014/


jacquelinekay:
+1

PR #160, opened 2016-05-06
Fix null terminator

mikaelarguedas:
This fixes the null terminator character check in strings

Relies on https://github.com/ros2/rmw_connext/pull/154
Connects to https://github.com/ros2/rclpy/issues/17


wjwwood:
817937be3d19075b7e400494e3125641da197c92 lgtm

PR #161, opened 2016-05-12
fix parsing with new uncrustify version

dirk-thomas:
Don't ask me why but otherwise the parser in uncrustify will :boom: http://ci.ros2.org/job/ci_linux/1290/testReport/junit/(root)/projectroot/uncrustify

Connect to ament/uncrustify#13


dirk-thomas:
Obsolete with the latest version of `uncrustify` from master.

PR #164, opened 2016-05-18
Don't call out to C++ typesupport for C services

jacquelinekay:
WIP at the moment, however feedback is welcome.

In particular I'm not sure if the approach I took for validating the type support identifiers is very good.

I also played around with making the C typesupport functions for messages "namespaced" and removing the "static" qualifiers them. I didn't see the reason for static qualifiers in the first place.

@wjwwood we discussed passing a `convert` function to the functions changed here (send/take_request/response), but I actually think that would be a bit too messy since the convert functions for C and C++ have different signatures--I think I would have to change the C++ `convert` functions to have type-erased `void *` arguments, rather const references which is what they are now. What do you think?


jacquelinekay:
CI:
http://ci.ros2.org/job/ci_linux/1334/
http://ci.ros2.org/job/ci_osx/1047/
http://ci.ros2.org/job/ci_windows/1351/

jacquelinekay:
Looking into warnings on OSX and an rcl test failure.

jacquelinekay:
http://ci.ros2.org/job/ci_osx/1053/badge/icon

wjwwood:
I'm starting to work on updating this pr as I did for the opensplice one.

jacquelinekay:
ok, just as a heads up, I'm debugging an issue with strings with Connext Static right now on the rcl parameters branch. I locally reverted 14dd974 to see if it's the cause of the problem. I don't think our changes will conflict, but I will wait on your changes before pushing mine.

wjwwood:
Ok, I pushed my changes to this branch as well. I'm running builds locally and I'll start CI after a bit.

wjwwood:
> I'm working on a compiler error fix in the message template.

Sorry, this branch was the one with compiler errors: 07063e8

jacquelinekay:
I resolved the issue with publishing C strings over services.

Basically, the `DDS_String_dup` has to happen once message for each publish or send request/response call. Both publish and send use convert_ros_to_dds, which calls String_dup. Previously only publish was calling dup, which caused services sending strings to fail. Then when I changed both publish and convert_ros_to_dds to use dup, then publishing strings failed. Changing publish not to dup and only convert_ros_to_dds fixes this.

jacquelinekay:
[Build Status](http://ci.ros2.org/job/ci_linux/1363/)

Another test with the most recent change.

wjwwood:
https://github.com/ros2/rmw_opensplice/pull/136 has been merged, but this one will wait until the warnings on Windows have been addressed.

wjwwood:
I rebased this, since it could not be merged.

jacquelinekay:
Thanks.

jacquelinekay:
Warnings, etc. should be fixed here: [Build Status](http://ci.ros2.org/job/ci_windows/1373/)

wjwwood:
Looks like something strange happened here. All the commits are in twice?

jacquelinekay:
there was some funny business with rebasing. I'm working on it...

jacquelinekay:
Not sure if the rebase fixed it. Want me to spend some more time cleaning up the commits? I plan to squash anyway.

jacquelinekay:
here's another (connext only) set of CI to ensure that this didn't break anything

[Build Status](http://ci.ros2.org/job/ci_linux/1378/)
[Build Status](http://ci.ros2.org/job/ci_osx/1075/)
[Build Status](http://ci.ros2.org/job/ci_windows/1384/)

rmw_connext_cpp/src/functions.cpp
```diff
@@ -1075,11 +1084,14 @@ rmw_create_service(
...
-    rosidl_typesupport_connext_cpp::typesupport_connext_identifier,
-    return NULL)
+
+  if (type_support->typesupport_identifier !=
+    rosidl_typesupport_connext_cpp::typesupport_connext_identifier &&
+    type_support->typesupport_identifier != rosidl_typesupport_connext_c__identifier)
+  {
+    RMW_SET_ERROR_MSG("type support identifier did not match valid typesupports");
+    return NULL;
+  }
```
dirk-thomas: A local macro would help to deduplicate this check.

rosidl_typesupport_connext_c/cmake/rosidl_typesupport_connext_c_generate_interfaces.cmake
```diff
@@ -164,11 +166,15 @@ if(WIN32)
   target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
     PRIVATE "ROSIDL_TYPESUPPORT_CONNEXT_C_BUILDING_DLL")
   target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
+    PRIVATE "ROSIDL_GENERATOR_C_${PROJECT_NAME}_BUILDING_DLL")
```
dirk-thomas: Where is this definition being used? I only see the one being set below (`ROSIDL_GENERATOR_C_BUILDING_DLL_${PROJECT_NAME}`).

jacquelinekay: hmm, I think this line is a relic of a bad rebase.

rosidl_typesupport_connext_c/cmake/rosidl_typesupport_connext_c_generate_interfaces.cmake
```diff
@@ -164,11 +166,15 @@ if(WIN32)
...
+  target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
     PRIVATE "NDDS_USER_DLL_EXPORT_${PROJECT_NAME}")
 endif()
 # The following still uses CPP because the Connext code which uses it was generated for CPP.
 target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
   PRIVATE "ROSIDL_TYPESUPPORT_CONNEXT_CPP_BUILDING_DLL_${PROJECT_NAME}")
+target_compile_definitions(${rosidl_generate_interfaces_TARGET}${_target_suffix}
+  PRIVATE "ROSIDL_GENERATOR_C_BUILDING_DLL_${PROJECT_NAME}")
```
dirk-thomas: Since this target is *not* building the `rosidl_generator_c` library I think this is highly dubious. Either some location where these macros are being used is wrong or the semantic of these definition is and should be changes.

jacquelinekay: This library defines a function which is forward-declared in a header provided by `rosidl_generator_c` with `ROSIDL_GENERATOR_C_PUBLIC_@(pkg_name)`:

https://github.com/ros2/rosidl/blob/master/rosidl_generator_c/resource/msg__type_support.h.em#L46-L49

I'm following the suggestion of @wjwwood, who designed the `ROSIDL_GET_TYPE_SUPPORT` mechanism. Maybe he can weigh in here.

rmw_connext_cpp/src/functions.cpp
jacquelinekay: fd3d155ccfc8d320024e47428958b70fc590590a implements the macro.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -46,13 +46,25 @@
...
 
+// Pass an object with a typesupport identifier member to compare
+#define RMW_CONNEXT_CHECK_TYPESUPPORT_IDENTIFIER(TYPESUPPORT) \
+  if (TYPESUPPORT->typesupport_identifier !=  /* NOLINT */ \
+    rosidl_typesupport_connext_cpp::typesupport_connext_identifier && \
+    TYPESUPPORT->typesupport_identifier != rosidl_typesupport_connext_c__identifier) \
+  { \
+    RMW_SET_ERROR_MSG("type support identifier did not match valid typesupports"); \
+    return NULL; \
+  } \
```
dirk-thomas: Shouldn't have a trailing backslash.

jacquelinekay: 9da657f

rosidl_typesupport_connext_c/cmake/rosidl_typesupport_connext_c_generate_interfaces.cmake
wjwwood: There is a function that is declared in the type support agnostic code that is implemented in this type support specific library. You can either pass the building dll def here or you can explicitly mark the header import and the def export and the functions that also call that function in this library need it forward declared to avoid the warnings Jackie had.

jacquelinekay:
[Build Status](http://ci.ros2.org/job/ci_windows/1393/)

alright, this should fix the inconsistent dll linkage warning on the identifier

jacquelinekay:
I would appreciate a re-review for this since I fixed the merge conflict and it's been nearly a month since I submitted it.

[Build Status](http://ci.ros2.org/job/ci_linux/1480/)
[Build Status](http://ci.ros2.org/job/ci_osx/1150/)
[Build Status](http://ci.ros2.org/job/ci_windows/1491/)

jacquelinekay:
Forgot to enable Connext in the first batch:

[Build Status](http://ci.ros2.org/job/ci_linux/1481/)
[Build Status](http://ci.ros2.org/job/ci_osx/1151/)
[Build Status](http://ci.ros2.org/job/ci_windows/1492/)

rosidl_typesupport_connext_c/resource/srv__type_support_c.cpp.em
```diff
@@ -34,9 +34,15 @@
 // is in the include/rosidl_typesupport_connext_cpp/impl folder
 #include "rosidl_generator_c/message_type_support.h"
 #include "rosidl_typesupport_connext_c/identifier.h"
-#include "rosidl_typesupport_connext_c/visibility_control.h"
 // Provides the definition of the service_type_support_callbacks_t struct.
 #include "rosidl_typesupport_connext_cpp/service_type_support.h"
+// Provides the definition of the message_type_support_callbacks_t struct.
+#include "rosidl_typesupport_connext_cpp/message_type_support.h"
```
dirk-thomas: Nitpick: should be two lines above for alphabetical order.

dirk-thomas:
Is there any code utilizing this and checking that it works as expected?

jacquelinekay:
Yes, the tests are all in the c_parameters branch. I can separate the tests into a standalone branch in rcl if you prefer.

jacquelinekay:
The existing tests for C services happened to work because the data fields were primitive types, but using C++ typesupport for strings and arrays when C typesupport should be used causes a segfault.

dirk-thomas:
So if the tests are in a differently named branch (`c_parameters`) then they are not being tested by the CI builds.

Also should the existing tests for C services be extended to cover strings and arrays to "prove" that this works?

jacquelinekay:
Yes, in my original post I asked if you would like me to make standalone tests for string and array types for C services. Since you are implying you won't accept this change without CI I will do so in a branch with the same name as this one.

I also suspect that any tests for string and array types for C services will fail for Connext Dynamic without #166.

dirk-thomas:
I am not implying anything. I simply can't judge if the code in this patch will work if it is not being exercised anywhere. From just looking at the diff it looks good to me. We can get there incrementally and merge this without any code using / testing it for now if that makes it easier.

jacquelinekay:
OK. The OSX and Windows CI jobs I launched are failing because of what looks like faulty configuration (permission denied errors). But if you approve I'm just going to merge this.

http://ci.ros2.org/job/ci_osx/1151/console#console-section-3
http://ci.ros2.org/job/ci_windows/1492/console

dirk-thomas:
A successfully build on OS X and Windows would be good before merging this.

jacquelinekay:
http://ci.ros2.org/job/ci_osx/1158/ (just started)
http://ci.ros2.org/job/ci_windows/1498/testReport/ (contains only Connext Dynamic failures, which I expected)

jacquelinekay:
the OSX job I linked to previously didn't have Connext enabled because we had to debug issues with OSX CI: http://ci.ros2.org/job/ci_osx/1164/

jacquelinekay:
I think the OSX job looks good except that it looks like Connext 5.2.3 introduces a TON of new warnings in Clang (due to message interfaces).

This seems to be a problem on master too.
http://ci.ros2.org/job/ci_osx/1175/warnings7Result/

dirk-thomas:
The problem has already been fixed on master: ros2/rmw_connext#179

jacquelinekay:
ok, thanks. Want me to rerun the OSX job or do you think this is ok to merge?

dirk-thomas:
The OS X job looks fine (only the warnings which have been addressed separately already). But the Windows build seems to have a least one new warning: http://ci.ros2.org/job/ci_windows/1498/warnings34Result/new/

Also there is still one comment pending.

jacquelinekay:
ah, damnit, I didn't catch that in my local Windows build because I didn't have Connext installed...

jacquelinekay:
I've spent hours trying to debug the inconsistent dll warning on Windows and I have no idea what to do about it. It follows the same pattern as `rosidl_typesupport_opensplice_c` except that instead of assigning the constant identifier to the C++ equivalent it assigns it to a constant C string. I could just go back to duplicating that identifier between C and C++ typesupport for Connext which would probably get rid of the warning, but I'd prefer not to do that since it may help catch errors. If you have any suggestions feel free to comment.

jacquelinekay:
I'm asking William for advice on the Windows warning. Otherwise the nitpick comment has been addressed and this pull request has been rebased.

wjwwood:
@jacquelinekay I rebased this (it had double commits and stuff) and squashed some of the commits.

I also pushed 4cbd50b which I believe should fix the warnings. Here is some CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1573/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1206/)
- Windows:
  - [Build Status](http://ci.ros2.org/job/ci_windows/1526/)

wjwwood:
Err... maybe 55da921.

jacquelinekay:
http://ci.ros2.org/job/ci_windows/1527/ then?

wjwwood:
Yeah.

jacquelinekay:
Thanks a ton for looking at that. Think this is ok to merge?

PR #165, opened 2016-05-20
support wait_for_service

wjwwood:
Connects to ros2/ros2#215

I'm currently dealing with an issue where the locally created DataReaders and DataWriters do not notify our listener: https://community.rti.com/forum-topic/built-publication-data-local-datawriters-and-datareaders


wjwwood:
I'm putting this one up for review as well. Some notes:

- The `rmw_service_server_is_available` function is 'not implemented' in this pr, since it will not answer the question if the service server is local.
- The "graph" guard condition will not wake up when local publishers, subscriptions, or service servers are created.

Related tests in `rcl` are disabled for Connext until those are resolved.

Both of these issues are related to the fact that Connext doesn't notify on local creation of DataWriter's and DataReader's, see: https://github.com/ros2/ros2/issues/215#issuecomment-220721174

I still think we should review and merge this as-is for two reasons: it is incrementally better and does work for different participants, and without the functions at least implemented rcl won't link.

After merging I'll create issues on this repository to track the remaining issues.

rosidl_typesupport_connext_cpp/resource/srv__type_support.hpp.em
```diff
@@ -83,6 +83,11 @@ send_response__@(spec.srv_name)(
   void * untyped_replier,
   const rmw_request_id_t * request_header,
   const void * untyped_ros_response);
+
+ROSIDL_TYPESUPPORT_CONNEXT_CPP_PUBLIC
+const char *
+get_request_topic_name__@(spec.srv_name)(void * untyped_requester);
```
tfoote: This change does not appear in the parallel change is that expected? https://github.com/ros2/rmw_opensplice/pull/137/files

tfoote:
lgtm

rosidl_typesupport_connext_cpp/resource/srv__type_support.hpp.em
wjwwood: Yeah this is expected, because in OpenSplice the "requester" object is available and I can just call `get_topic_name` (or something like that) on it. In Connext, the requester object is type erased, so I have to ask the generated code to cast it to the right type and then call that function on it.

PR #169, opened 2016-06-03
fix bounded arrays

dirk-thomas:
At least compiling the generated code for unbounded array passes with this.

Connect to ros2/system_tests#123.


dirk-thomas:
Please review.

mikaelarguedas:
lgtm

PR #171, opened 2016-06-07
use Connext 5.2.3

dirk-thomas:
The latest version also comes with `x64Darwin15clang7.0` but `x64Darwin14clang6.0` should still work.

Since our code doesn't compile with Visual Studio 2013 anymore we can ignore these locations now.

Connect to ros2/ros2#232.


dirk-thomas:
Please review.

tfoote:
+1

dirk-thomas:
Fixed wrong VS2015 logic in 0c4cf3dd64777e20f53b67be6c1f5ff97c98f024.

PR #172, opened 2016-06-08
use rtiddsgen_server.bat on Windows if available

dirk-thomas:
The generation for an average number of messages is about half with the patch.

Before: http://ci.ros2.org/job/ci_windows/1433/
After: http://ci.ros2.org/job/ci_windows/1431/


dirk-thomas:
Please review.

wjwwood:
+1

PR #175, opened 2016-06-15
flag to switch between ABI versions of libstdc++

dirk-thomas:
The binary packages provided by RTI atm are not using the C++11 ABI but the old ABI (see https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html). So for building libraries on top of e.g. `connext_cpp_requester_details` the following must be set (as before): `#define _GLIBCXX_USE_CXX11_ABI 0`. The same is the case for the 5.2.0 Debian package created on Trusty.

In that case `connext_cmake_module` is exporting `Connext_GLIBCXX_USE_CXX11_ABI_ZERO` which will result in `#define _GLIBCXX_USE_CXX11_ABI 0` to be used (same as before: http://ci.ros2.org/job/ci_linux/1449/ (I already addressed the missing copyright))

But in the case of the newly build Connext 5.2.3 package (on Xenial with G++ 5.2.3) the define must not be set (we don't have CI to test this).

### DIFF
--- a/connext_cmake_module/CMakeLists.txt
+++ b/connext_cmake_module/CMakeLists.txt
@@ -4,6 +4,7 @@ project(connext_cmake_module)
 
 find_package(ament_cmake REQUIRED)
 
+set(${PROJECT_NAME}_DIR "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 list(INSERT CMAKE_MODULE_PATH 0 "${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules")
 find_package(Connext MODULE)
 if(Connext_HOME)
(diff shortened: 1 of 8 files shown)


jacquelinekay:
Interesting approach, I did not know about `try_compile` in CMake. Why can't we drop support for the binary package provided by RTI that does not use the C++11 ABI if we have our own 5.2.3 package on Xenial?

dirk-thomas:
Because we want users to be able to compile the code too.

jacquelinekay:
ah, right. So RTI doesn't offer a binary of 5.2.3 yet?

dirk-thomas:
They do, but they build their binary packages with gcc 4.8.

dirk-thomas:
Please review this.

wjwwood:
lgtm

PR #177, opened 2016-06-22
use target_compile_definitions instead of add_definitions for try compile

dirk-thomas:
Connect to ros2/rcl#66.

http://ci.ros2.org/job/ci_linux/1506/


dhood:
+1

PR #173, opened 2016-06-08
export --no-as-needed link flag

dirk-thomas:
Export the linker flags for the Connext packages to address https://github.com/ros2/rcl/pull/55#issuecomment-224669618 (as documented in the [Connext Platform notes](https://community.rti.com/static/documentation/connext-dds/5.2.3/doc/manuals/connext_dds/RTI_ConnextDDS_CoreLibraries_PlatformNotes.pdf)).

http://ci.ros2.org/job/ci_linux/1419/

PR #178, opened 2016-06-22
Use --no-as-needed in abi check

dhood:
can be changed to `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} @Connext_LINK_FLAGS@")` after https://github.com/ros2/rmw_connext/pull/173 is merged

connects to https://github.com/ros2/rcl/issues/66

### DIFF
--- a/connext_cmake_module/cmake/check_abi.cmake
+++ b/connext_cmake_module/cmake/check_abi.cmake
@@ -14,7 +14,7 @@
 
 cmake_minimum_required(VERSION 2.8.3)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11 -Wall -Wextra")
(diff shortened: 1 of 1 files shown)


connext_cmake_module/cmake/check_abi.cmake
```diff
@@ -21,3 +21,4 @@ include_directories(@Connext_INCLUDE_DIRS@)
 add_executable(exe
   "@connext_cmake_module_DIR@/check_abi.cpp")
 target_link_libraries(exe @Connext_LIBRARIES@)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wl,--no-as-needed")
```
dirk-thomas: Please append the new flags to the already existing line 17.

dirk-thomas:
+1

PR #180, opened 2016-06-22
support bounded vectors

dirk-thomas:
Connect to ros2/rosidl#158.

I will comment on diff for important changes.

### DIFF
--- a/rosidl_typesupport_connext_c/resource/msg__type_support_c.cpp.em
+++ b/rosidl_typesupport_connext_c/resource/msg__type_support_c.cpp.em
@@ -182,7 +182,7 @@ convert_ros_to_dds(const void * untyped_ros_message, void * untyped_dds_message)
       )()->data);
 @[  end if]@
 @[  if field.type.is_array]@
-@[    if field.type.array_size]@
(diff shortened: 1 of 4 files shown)


rosidl_typesupport_connext_c/resource/msg__type_support_c.cpp.em
```diff
@@ -182,16 +182,31 @@ convert_ros_to_dds(const void * untyped_ros_message, void * untyped_dds_message)
...
       return false;
     }
+@[      if field.type.is_upper_bound]@
+    if (size > @(field.type.array_size)) {
+      fprintf(stderr, "array size exceeds upper bound\n");
+      return false;
+    }
+@[      end if]@
     DDS_Long length = static_cast<DDS_Long>(size);
-    dds_message->@(field.name)_.ensure_length(length, length);
```
dirk-thomas: Before this was allowing to resize the sequence to an arbitrary length.

dirk-thomas: The return value must also be checked if the request was successful.

wjwwood:
lgtm

PR #182, opened 2016-06-24
fix string comparison since variable might not be defined

dirk-thomas:
Regression of #176.

Since the variable might not be defined the comparison needs to explicitly evaluate the variable.


tfoote:
+1

Issue #181, opened 2016-06-24
support reliable large data publishing with asynchronous publisher


wjwwood:
When trying to publish large messages reliably with Connext (like images with programs in `image_tools`) we get a message like this:

```
COMMENDSrWriterService_write:!write. Reliable large data requires asynchronous writer.
```

See: https://community.rti.com/examples/asynchronous-publisher

PR #183, opened 2016-06-25
enable async publish mode for publishers

wjwwood:
Fixes https://github.com/ros2/rmw_connext/issues/181

I'd like feedback on whether or not we should always turn this on. I'm guessing there is a performance impact for turning it on, and so it may not be wise to do so. As I see it, we have a few options:

- turn it on always (this pr)
- expose an option for our users, throw an exception when the data is too large and the option is off (connext's behavior)
- use a heuristic to determine when to turn it on, but default to off

The last option is clearly attractive since it "just works" and would save performance when it could, but I see two problems with the approach: figuring out a heuristic that always works for Connext and differences in behavior when this is enabled vs disabled. It would be surprising for the publish function to change behavior depending on the message you're using.

Also, I'm pretty sure you cannot change this setting after creating the publisher, so we'll have to have a heuristic that can determine whether or not to turn it on before starting to publish data.

My first guess would be something like, turn on async publishing:

- If reliable and
  - if message has any unbounded fields (potentially infinite size) or
  - if message has all bounded fields, but the fixed max size is too big (whatever Connext's cut off is)

The problem being that this would mean we have to turn on async publishing for pretty much all of our messages, since many have unbounded strings.

For now, we can just turn it on always and I can ticket the need to figure out when to turn it off for performance gains.

### DIFF
--- a/rmw_connext_shared_cpp/src/shared_functions.cpp
+++ b/rmw_connext_shared_cpp/src/shared_functions.cpp
@@ -220,6 +220,8 @@ get_datawriter_qos(
     return false;
   }
 
+  // TODO(wjwwood): conditionally use the async publish mode using a heuristic:
+  //  https://github.com/ros2/rmw_connext/issues/190
   datawriter_qos.publish_mode.kind = DDS::ASYNCHRONOUS_PUBLISH_MODE_QOS;
 
   return true;
--- a/rmw_connext_shared_cpp/src/shared_functions.cpp
+++ b/rmw_connext_shared_cpp/src/shared_functions.cpp
@@ -220,6 +220,8 @@ get_datawriter_qos(
     return false;
   }
 
+  datawriter_qos.publish_mode.kind = DDS::ASYNCHRONOUS_PUBLISH_MODE_QOS;
+
   return true;
 }
 


wjwwood:
CI:

- Linux:
  - [Build Status](http://ci.ros2.org/job/ci_linux/1556/)
- OS X:
  - [Build Status](http://ci.ros2.org/job/ci_osx/1202/)

dirk-thomas:
I would expose a flag in the API to turn it on or off. The default should be using some heuristic (as you described: bounded size below threshold).

wjwwood:
tl;dr I'd like to merge as-is and ticket parameterization and optimization. What do others think?

----

My issue with exposing the option right now is that it isn't clear to me how it should work for middlewares which don't have this option. I feel that it's exposing some implementation specific details up to the user level. Ideally we wouldn't do that because it either makes the option complicated (what does it do on systems which do not have "async publishing") or it couples our interface to particular kinds of implementations (we require the notion of asynchronous publishing to implement rmw).

If we add an parameter to control this, then I think we need to make it a ROS level general concept that every implementation needs to support. I think it is potentially related to the need for us to clarify our blocking behavior for publish, see: https://github.com/ros2/rcl/blob/master/rcl/include/rcl/publisher.h#L173-L175

My preference would be to turn it on always for now, as that will "just work", in order to fix some tests and demos that currently won't work with Connext. This would be at the expense of some performance, but just until we figure out how to parameterize it or optimize it correctly. Admittedly, that might take a while to address.

Btw, this applies to FastRTPS too, because it also has this setting (which is an extension of the DDS spec).

wjwwood:
Bump, @ros2/ros2_team can you comment on this when you all get time?

wjwwood:
Ok, I'll ticket the issues of having a better heuristic and having the options exposed through the rcl api.

wjwwood:
I ticketed both issues:

- https://github.com/ros2/rmw_connext/issues/190
- https://github.com/ros2/ros2/issues/255

With clean CI (as far as I can tell) and a +1 from @codebot, I'll merge this so that we have large message publishing for the image demos.

PR #184, opened 2016-06-28
use correct ABI for C typesupport if not C++11

dirk-thomas:
This is the same logic as for the C++ type support. That makes it compile for me against the RTI binary again.


dhood:
+1 fixes linking errors when using connext off the website

PR #185, opened 2016-06-29
retry Connext code generation on Windows if the first invocation does nothing

dirk-thomas:
While I can't reproduce the problem in a minimal example it happened frequently (only on Windows) in my recent CI builds. In the case the generator doesn't generate the expected files it also doesn't output anything to `stdout` or `stderr` and has a return code of `0`.

I suspect this to be a race condition in `rtiddsgen_server` (since it doesn't happen with `rtiddsgen`).

E.g. http://ci.ros2.org/job/ci_windows/1564/consoleFull (search for "retrying once on Windows")

Please review.

### DIFF
--- a/rosidl_typesupport_connext_cpp/rosidl_typesupport_connext_cpp/__init__.py
+++ b/rosidl_typesupport_connext_cpp/rosidl_typesupport_connext_cpp/__init__.py
@@ -14,6 +14,7 @@
 
 import os
 import subprocess
+import sys
 
 from rosidl_cmake import convert_camel_case_to_lower_case_underscore
 from rosidl_cmake import expand_template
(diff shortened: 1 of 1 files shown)


mikaelarguedas:
LGTM,
That's unfortunate that we have to check the files' existence ourselves, but I guess we'll have to do it until it's fixed by RTI.

dirk-thomas:
After more builds it also happened on OS X, so retrying on all platforms now: ce056f8ba0f0e90f0a29deb87ee13e0c0a9f78f1

dirk-thomas:
And sometimes more than one retry is necessary: fe273cb673f8d86dd1be1fa4f081bcb709864c51

PR #166, opened 2016-05-24
replace much of the macros in connext dynamic with templates

jacquelinekay:
Still in progress.

PR #187, opened 2016-06-29
Replace most of the macros in rmw_connext_dynamic_cpp with templates

jacquelinekay:
#166 with heavily rewritten history, as requested by @dirk-thomas. I started a new branch because I was paranoid.


jacquelinekay:
[Build Status](http://ci.ros2.org/job/ci_linux/1600/)
[Build Status](http://ci.ros2.org/job/ci_osx/1214/)

dirk-thomas:
You won't like the refactoring from #189 because it moves all `publish` and `take` related functionality into a separate compilation unit. Sorry.

jacquelinekay:
So, do you want to close this one, do you want to resolve it after #189 is reviewed and merged, or do you want to review this one and then review and merge #189?

dirk-thomas:
Since #189 fixes a huge problem I think it has priority. We can rebase this afterwards while maintaining the separate compilation units. But I think it would make sense to also split the functions / templates as in `rmw_opensplice` in order to not have a few monolithic files with thousands of lines.

wjwwood:
I'd say merge this one first and refactor #189, since it was open first and appears ready for review/merge. Also, in my estimation, it will be simpler to refactor #189 on top of this merged than the other way around, because #189 is *mostly* just moving code, whereas this one is also heavily changing the code.

dirk-thomas:
Moving the code into headers (as in this PR) doesn't work for #189 since it must be split in two separate compilation units.

wjwwood:
Why not? Can't #189 then create two headers and two cpp files and include one header in each? Seems to me anything that was previously a macro in a cpp file and split into two can still be done after this refactoring, in principal. Maybe you can give a more specific example of where it will not work (a particular function or macro that's now a template)?

dirk-thomas:
That would be possible but very fragile. The definition in the `.cpp` file determines the ABI being used. If the header files are included differently in the future this would break again. Therefore I think the code which has specific ABI constraints should stay in `.cpp` file to prevent this.

wjwwood:
Could you move the non-macro / non-template functions that are ABI sensitive to the separate file and call them from the macros / templates.

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -497,8 +255,6 @@ DDS_TypeCode * _create_type_code(
   return NULL;
 }
 
-extern "C"
-{
```
wjwwood: I was suspicious of this initially, but it actually two `extern "C"` sections getting merged.

rmw_connext_dynamic_cpp/src/templates.hpp
```diff
@@ -0,0 +1,1554 @@
...
+// multiple definitions of ambiguous primitive types
+SPECIALIZE_GENERIC_C_ARRAY(String, rosidl_generator_c__String)
+SPECIALIZE_GENERIC_C_ARRAY(bool, bool)
+SPECIALIZE_GENERIC_C_ARRAY(byte, uint8_t)
+SPECIALIZE_GENERIC_C_ARRAY(char, char)
+SPECIALIZE_GENERIC_C_ARRAY(float32, float)
+SPECIALIZE_GENERIC_C_ARRAY(float64, double);
+SPECIALIZE_GENERIC_C_ARRAY(int8, int8_t)
+// TODO(jacquelinekay) this results in an ambiguous definition
+// SPECIALIZE_GENERIC_C_ARRAY(uint8, uint8_t)
```
wjwwood: Any ideas on what you wanted to do here @jacquelinekay? It's a tough one. I don't know if you can have a template specialization with typdef'ed or aliased types and avoid the redefinition issue. On the other hand the structures and functions for byte and uint8_t in `rosidl_generator_c` are basically identical. Maybe we should do nothing and put a comment explaining why?

wjwwood:
Other than the TODO I commented on, the code changes look good to me. I think it might be prudent to figure out what we want to do about the TODO before merging though. @jacquelinekay do you have any ideas? We could put the type in a structure which could then be templated correctly, but that seems to be working against the goal of reducing code duplication.

rmw_connext_dynamic_cpp/src/functions.cpp
jacquelinekay: Yup.

rmw_connext_dynamic_cpp/src/templates.hpp
jacquelinekay: Hmm. Uncommenting this results in a compilation error which is why I originally wrote that TODO. The specialization that I commented out is unnecessary right now. A uint8_t will always map to a "byte" type in DDS. There's no way of differentiating whether a uint8_t should map to a "byte" or a "uint8" in DDS without attaching some extra context to the type. Just curious, why do we distinguish between "byte" and "uint8"? Is it because it's ambiguous whether "byte" is signed or unsigned?

wjwwood: I think it's a legacy of ROS 1 that we have both `byte` and `uint8`. The structures and functions are differently named, but aside from that are identical. It might "just work", but if anything were ever different between byte and uint8 in our code, it would be a problem.

jacquelinekay: OK, I think I can rework this to use the type_id defined by rosidl_typesupport_introspection to resolve the ambiguity.

jacquelinekay: Well, the fix I originally had in mind doesn't work because rosidl_typesupport_introspection_c__MessageMember is not a constexpr (and can't be because it has C linkage). trying a different approach...

wjwwood: The other option is to avoid using the templated version of GenericArray, which isn't as nice, but it works around the issue fairly well. I'm not sure how much of a complication that is though.

jacquelinekay:
Just rebased against master.

jacquelinekay:
I think d1a73b5 will fix potential future issues with byte vs. uint8 by being more exact about matching the ROS ids with the rosidl_generator_c and pure C types.

jacquelinekay:
I'm testing the changes locally by running linters and test_rclcpp tests. I'll merge if it looks good.

May need to apply a similar change to rmw_fastrtps_cpp as I used a similar technique there to get introspection_c typesupport.

rmw_connext_dynamic_cpp/src/macros.hpp
```diff
@@ -18,13 +18,94 @@
...
+      values, \
+      &length, \
+      NULL, \
+      static_cast<DDS_DynamicDataMemberId>(index)); \
+  } \
 
-#define C_STRING_ASSIGN(A, B) rosidl_generator_c__String__assignn(&A, B, size - 1);
 
-#define CPP_STRING_ASSIGN(A, B) A = B;
+#define SPECIALIZE_GENERIC_C_ARRAY(ROS_ID, C_NAME, C_TYPE) \
```
dirk-thomas: Please use a different name. There is not term "ROS Id". Maybe `ROS_TYPE_ID`?

wjwwood:
d1a73b5 looks ok to me.

jacquelinekay:
tests passed, I addressed some of the most recent nitpicks, merging now.

dirk-thomas:
I guess the new failing tests are related to this merge (http://ci.ros2.org/job/ci_linux/1632/). Please address those.

jacquelinekay:
#191

PR #188, opened 2016-07-01
wrap Connext libraries in --no-as-needed

dirk-thomas:
Replaces #173, requires ament/ament_cmake#75.

### DIFF
--- a/connext_cmake_module/cmake/Modules/FindConnext.cmake
+++ b/connext_cmake_module/cmake/Modules/FindConnext.cmake
@@ -305,6 +305,8 @@ if(Connext_FOUND)
   endif()
 
   if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    set(Connext_LIBRARIES "-Wl,--no-as-needed" ${Connext_LIBRARIES} "-Wl,--as-needed")
+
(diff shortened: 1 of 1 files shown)


wjwwood:
I'm trying out a build on my Linux VM with this pr and https://github.com/ament/ament_cmake/pull/75, but do I need https://github.com/ros2/rcl/pull/61 too?

dirk-thomas:
No, the `export_link_flags` is unrelated and should not be mixed.

With these two patches it should compile and link but the runtime behavior of Connext dynamic is still broken. Testing a patch for that right now.

dirk-thomas:
At least it doesn't seem to break anything with the normal builds:

* http://ci.ros2.org/job/ci_linux/1616/
* http://ci.ros2.org/job/ci_osx/1220/
* http://ci.ros2.org/job/ci_windows/1581/

wjwwood:
I'm retrying the build on my VM now.

wjwwood:
Works for me.

dirk-thomas:
Does that imply a +1 on the ament_cmake PR too?

wjwwood:
Yeah, I think it's good to merge.

PR #189, opened 2016-07-01
build rmw_connext_dynamic_cpp in two parts to match ABI of Connext as well as ROS

dirk-thomas:
While ROS 2 packages on Xenial are always built with the new ABI (`_GLIBCXX_USE_CXX11_ABI 1`) Connext might use either or. This patch moves all API which interacts with the ROS messages(namely the `std::string` fields) into a separate compilation unit which is always build with the new ABI. The remaining `rmw` functions interacting with Connext are built with the same ABI as Connext. Since the interface between the two compilation units doesn't contain anything which is different between the two ABIs it doesn't pose a problem that both are built with a different ABI.

* No change with the custom Debian package of Connext: http://ci.ros2.org/job/ci_linux/1624/
* Also no change on OS X and Windows:
  * http://ci.ros2.org/job/ci_osx/1223/
  * http://ci.ros2.org/job/ci_windows/1590/
* And the dynamic tests are passing with the official Connext binaries now:
  * http://ci.ros2.org/job/ci_linux/1625/ (before http://ci.ros2.org/job/ci_linux/1620/)

The code related to `publish` and `take` has only been moved to the separate compilation unit but has not been modified in any way.

Please review.


wjwwood:
lgtm, but I still feel that merging https://github.com/ros2/rmw_connext/pull/187 first would be easiest.

dirk-thomas:
Then please go ahead with merging it.

dirk-thomas:
Closing in favor of #191.

PR #191, opened 2016-07-06
Fix linter errors

jacquelinekay:


jacquelinekay:
Here's a Connext Dynamic-only set of CI jobs:

[Build Status](http://ci.ros2.org/job/ci_linux/1633/)
[Build Status](http://ci.ros2.org/job/ci_osx/1227/)

rmw_connext_dynamic_cpp/src/macros.hpp
```diff
@@ -84,7 +86,7 @@
     } \
   }; \
   template<> \
-  struct IdTypeMap<rosidl_typesupport_introspection_c__ROS_TYPE_ ## ROS_TYPE_ID> { \
+  struct IdTypeMap<rosidl_typesupport_introspection_c__ROS_TYPE_ ## ROS_TYPE_ID>{ \
```
wjwwood: hrm, that's annoying...

wjwwood:
code changes lgtm

wjwwood:
I'm going to merge this to get rid of the lint failures on master.

jacquelinekay:
I was looking into some connext dynamic failures in those CI jobs. But I'll try to fix them in a subsequent pr.

wjwwood:
Oh, sorry, I didn't realize. I think this was a strict improvement though right?

jacquelinekay:
No, based on some local testing with different states before and after this change, I think I introduced a subtle bug here: https://github.com/ros2/rmw_connext/pull/191/files#diff-a1342c2267183e68decd53f4e4273732R1239

I'll have another patch in a sec.

PR #192, opened 2016-07-06
build rmw_connext_dynamic_cpp in two parts to match ABI of Connext as well as ROS

wjwwood:
Rebase and fixup of https://github.com/ros2/rmw_connext/pull/189 since the merge of #187.

@dirk-thomas I think this branch is ok, but you should look it over closely. It builds on my mac, but I haven't re-tested it on CI or in my Linux VM for the ABI problem, though if I understand the strategy, I think this should work.

Like your branch, this pr only moves code around and is careful about what functions get included where (either the functions.cpp file or the publish_take.cpp file, which have different ABI settings).


dirk-thomas:
Seems to work with the official RTI binaries which use the old ABI: http://ci.ros2.org/job/ci_linux/1636/#showFailuresLink

+1

wjwwood:
I'll merge this then.

dirk-thomas:
Linter error: http://ci.ros2.org/job/ci_linux/1643/testReport/junit/(root)/projectroot/cpplint/

dirk-thomas:
Also these failing tests are new - not sure if from this change or from the refactoring to templates (@jacquelinekay): http://ci.ros2.org/job/ci_linux/1643/testReport/junit/(root)/test_publisher_subscriber_cpp__BoundedArrayPrimitives__rmw_connext_dynamic_cpp_/test_publisher_subscriber/

jacquelinekay:
I believ #195 will fix the failing tests.

Issue #168, opened 2016-05-28
detect participant-local graph changes


wjwwood:
For reference:

- https://github.com/ros2/ros2/issues/215#issuecomment-221745738
- https://community.rti.com/forum-topic/built-publication-data-local-datawriters-and-datareaders

Basically, any newly created DataWriters and DataReaders generate an entry on a "builtin" DDS topic. In OpenSplice you get all notifications, but Connext follows a section of the DDS spec that says locally created (created in the same participant) DataWriters and DataReaders don't generate entries (the specifics are in the above links).

So for us to get notifications of local changes, we'll need to maintain some state ourselves.

Because of this I disabled the `rcl_service_server_is_available` function for Connext and Connext Dynamic:

- https://github.com/ros2/rmw_connext/blob/c983a8c592b094c3c3dd98a52d4d24c69e65840f/rmw_connext_dynamic_cpp/src/functions.cpp#L3327-L3329
- https://github.com/ros2/rmw_connext/blob/c983a8c592b094c3c3dd98a52d4d24c69e65840f/rmw_connext_dynamic_cpp/src/functions.cpp#L3327-L3329

I also disabled the related tests in `rcl` (which should be re-enabled after they're fixed):

- https://github.com/ros2/rcl/blob/818b15463b736e2198a47b2818b2630a59afbe12/rcl/test/rcl/test_node.cpp#L247-L250
- https://github.com/ros2/rcl/blob/818b15463b736e2198a47b2818b2630a59afbe12/rcl/test/CMakeLists.txt#L70-L71

dhood:
more disabled related tests in `rclcpp_examples`:
https://github.com/ros2/examples/blob/035c2e923dbecf0bb0238b7b9887b354ffb0af10/rclcpp_examples/CMakeLists.txt#L79

dirk-thomas:
Once this is merged for Connext as well as FastRTPS the conditional checks last updated in ros2/system_tests#145 can be removed.

wjwwood:
Just an update on this. I have a solution that "works". However, there are still spurious failures which I believe occur due to a race condition.

[...]

PR #194, opened 2016-07-06
detect and report local graph changes

wjwwood:
Fixes https://github.com/ros2/rmw_connext/issues/168

This required, apparently for the first time, that the node is given to the `rmw_destroy_client` and `rmw_destroy_service` calls, so that will change the rmw API.

I'm still debugging some issues with the tests.

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -1000,7 +1000,7 @@ fail:
 }
 
 rmw_ret_t
-rmw_destroy_client(rmw_client_t * client)
+rmw_destroy_client(rmw_node_t * node, rmw_client_t * client)
 {
   if (!client) {
     RMW_SET_ERROR_MSG("client handle is null");
@@ -1118,7 +1118,7 @@ rmw_create_service(
     return nullptr;
   }
 
-  auto node_info = static_cast<ConnextNodeInfo *>(node->data);
+  ConnextNodeInfo * node_info = static_cast<ConnextNodeInfo *>(node->data);
   if (!node_info) {
     RMW_SET_ERROR_MSG("node info handle is null");
     return NULL;
@@ -1235,7 +1235,7 @@ fail:
 }
 
 rmw_ret_t
-rmw_destroy_service(rmw_service_t * service)
+rmw_destroy_service(rmw_node_t * node, rmw_service_t * service)
 {
   if (!service) {
     RMW_SET_ERROR_MSG("service handle is null");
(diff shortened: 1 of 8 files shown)


wjwwood:
So, it seems that Connext's request-reply uses some strange topic names. For example, the service server creates:

```
+S add_two_intsRequest <example_interfaces::srv::dds_::AddTwoInts_Request_>
+P add_two_intsReply <example_interfaces::srv::dds_::AddTwoInts_Response_>
```

But the client creates:

```
+S add_two_intsReply_ac1702d4000020e70000000180002003 <example_interfaces::srv::dds_::AddTwoInts_Response_>
+P add_two_intsRequest <example_interfaces::srv::dds_::AddTwoInts_Request_>
```

Note what looks like a GUID attached to the reply topic. It's not clear to me how the replier (service server) delivers the reply to the requester (service client).

I spent the better part of last night trying to figure out a better way to figure out when the service server is available for how Connext implements Request-Reply. I have some ideas, but as of right now this is not ready for merging.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -176,7 +198,7 @@ rmw_create_publisher(
     return nullptr;
   }
 
-  auto node_info = static_cast<ConnextNodeInfo *>(node->data);
+  ConnextNodeInfo * node_info = static_cast<ConnextNodeInfo *>(node->data);
```
wjwwood: I think I did that so my autocomplete would work, but I can change them back: 90d4c86

dirk-thomas:
This recent change broke `rmw_connext_dynamic_cpp`: http://ci.ros2.org/job/ci_linux/1870/testReport/

While not being actively tested at the moment maybe it can be easily fixed @wjwwood ?

wjwwood:
I believe it is now failing because the wait_for_service is still not implemented, see:

http://ci.ros2.org/job/ci_linux/1870/testReport/(root)/projectroot/requester_replier_cpp__Primitives__rmw_connext_dynamic_cpp/

The tests could be disabled, but I don't know if it is worth it since we're not regularly trying to figure out which failures are known and which are not for that rmw implementation. The fact that this function is not implemented is already tracked in an issue: https://github.com/ros2/rmw_connext/issues/203

I'm not sure about the other failures.

dirk-thomas:
Since the merge numerous very simple tests which only do pub/sub are failing. I don't think the referenced ticket about `wait_for_service` is related to all of them. When I tried it with the new composition demos the executable crashed when calling one of the new functions around `add_information`. Therefore I am pretty sure that this is a regression of this PR.

wjwwood:
Well I'm sorry it is broken, because it wasn't my intention to break it, in fact I spent some extra time trying to keep it up to speed. However, fixing this is not at the top of my priority list because it's not  one of our supported rmw implementations right now.

PR #195, opened 2016-07-07
Correct null pointer checks

jacquelinekay:
Should fix the test_communication failures in connext dynamic.

[Build Status](http://ci.ros2.org/job/ci_linux/)
[Build Status](http://ci.ros2.org/job/ci_osx/1235/)

### DIFF
--- a/rmw_connext_dynamic_cpp/src/templates.hpp
+++ b/rmw_connext_dynamic_cpp/src/templates.hpp
@@ -520,12 +520,12 @@ bool set_value_with_different_types(
   T * ros_values = nullptr;
   if (member->is_array_) {
     size_t array_size = set_array_size_and_values<Id>(member, ros_message, ros_values);
-    if (!ros_values) {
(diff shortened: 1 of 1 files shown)


rmw_connext_dynamic_cpp/src/templates.hpp
```diff
@@ -637,7 +637,7 @@ bool set_value<rosidl_typesupport_introspection_cpp::ROS_TYPE_STRING>(
     size_t array_size =
       set_array_size_and_values<rosidl_typesupport_introspection_cpp::ROS_TYPE_STRING>(
       member, ros_message, ros_values);
-    if (!ros_values) {
+    if (!ros_values && array_size) {
```
tfoote: `array_size > 0` like the others lines?

jacquelinekay: Yeah, this will logically work because if array_size is not equal to zero it will evaluate to true in this conditional, but I'll change it to > 0 to be more readable.

dirk-thomas:
lgtm

PR #210, opened 2016-12-14
Ignore connext dynamic package while support is paused

dhood:
otherwise you get build errors


dirk-thomas:
This makes it impossible to "enable" `rmw_connext_dynamic_cpp` on the build farm. Therefore I would suggest to create that file locally for now. Even if that is more effort.

dhood:
ok we can add it to the source build instructions. alternatively it could just be present in the "beta" tag

dirk-thomas:
Adding it just for the tagging and removing it directly afterwards sounds good me.

dirk-thomas:
Instead of temporary ignoring this package #211 simply makes it compile again.

wjwwood:
I think it would still be good to disable it in the beta branch to reduce the compile time for people building from source (the additionally generated message stuff and libraries/executables adds to the build time considerably).

dirk-thomas:
Ok, we will add the ignore file just before beta and remove it immediately after tagging.

PR #211, opened 2016-12-16
fix compile errors with rmw_connext_dynamic_cpp

dirk-thomas:
Replaces #210.

### DIFF
--- a/rmw_connext_dynamic_cpp/src/functions.cpp
+++ b/rmw_connext_dynamic_cpp/src/functions.cpp
@@ -50,9 +50,6 @@
 #include "rmw/error_handling.h"
 #include "rmw/rmw.h"
 #include "rmw/types.h"
-// This header is in the rosidl_typesupport_connext_cpp package and
-// is in the include/rosidl_typesupport_connext_cpp/impl folder.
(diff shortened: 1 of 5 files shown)


rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -52,7 +52,7 @@
 #include "rmw/types.h"
 // This header is in the rosidl_typesupport_connext_cpp package and
 // is in the include/rosidl_typesupport_connext_cpp/impl folder.
```
Karsten1987: is that comment still valid?

dirk-thomas:
It currently still fails later on downstream packages - will get that sorted out soon...

rmw_connext_dynamic_cpp/src/functions.cpp
dirk-thomas: No, will remove the comment.

dirk-thomas:
Ready for review.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/2082/)
* Mac OS [Build Status](http://ci.ros2.org/job/ci_osx/1608/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/2100/)

PR #217, opened 2017-01-26
Rosnode list

Karsten1987:
Connect to ros2/cli_tools#1


Karsten1987:
There is one change for connext:
https://github.com/ros2/rmw_connext/pull/217/files#diff-9bba412d4597100825da2b2307403d08R322

We actively have to assign the name to the dds participant.

PR #219, opened 2017-03-02
Use -Wpedantic

mikaelarguedas:
connects to https://github.com/ros2/ros2/issues/305

### DIFF
--- a/rmw_connext_dynamic_cpp/src/macros.hpp
+++ b/rmw_connext_dynamic_cpp/src/macros.hpp
@@ -51,11 +51,11 @@
         "type support handle implementation '%s' (%p) does not match valid type supports " \
         "('%s' (%p), '%s' (%p))", \
         TYPE_SUPPORTS->typesupport_identifier, \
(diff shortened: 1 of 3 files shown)


rmw_connext_dynamic_cpp/src/macros.hpp
```diff
@@ -51,11 +51,11 @@
         "type support handle implementation '%s' (%p) does not match valid type supports " \
         "('%s' (%p), '%s' (%p))", \
         TYPE_SUPPORTS->typesupport_identifier, \
-        TYPE_SUPPORTS->typesupport_identifier, \
-        rosidl_typesupport_introspection_cpp::typesupport_identifier, \
+        reinterpret_cast<const void *>(TYPE_SUPPORTS->typesupport_identifier), \
```
dirk-thomas: Isn't a `static_cast` sufficient here?

Same below.

mikaelarguedas: good point, fixed for connext and connext dynamic in https://github.com/ros2/rmw_connext/pull/219/commits/9c43edf6200347f2bff1983a0e5648d01910b10d

PR #220, opened 2017-03-29
update to support node namespaces in rmw API

wjwwood:
Connects to ros2/rmw#95

### DIFF
--- a/rmw_connext_shared_cpp/src/shared_functions.cpp
+++ b/rmw_connext_shared_cpp/src/shared_functions.cpp
@@ -302,7 +302,11 @@ destroy_node_names(
 }
 
 rmw_node_t *
-create_node(const char * implementation_identifier, const char * name, size_t domain_id)
+create_node(
+  const char * implementation_identifier,
+  const char * name,
+  const char * name_space,
+  size_t domain_id)
 {
   DDSDomainParticipantFactory * dpf_ = DDSDomainParticipantFactory::get_instance();
   if (!dpf_) {
@@ -319,6 +323,10 @@ create_node(const char * implementation_identifier, const char * name, size_t do
   }
   participant_qos.participant_name.name =
     static_cast<char *>(rmw_allocate((strlen(name) + 1) * sizeof(char)));
+  if (!participant_qos.participant_name.name) {
+    RMW_SET_ERROR_MSG("failed to allocate memory");
+    return NULL;
+  }
   participant_qos.participant_name.name = strdup(name);
   // forces local traffic to be sent over loopback,
   // even if a more efficient transport (such as shared memory) is installed
@@ -430,6 +438,14 @@ create_node(const char * implementation_identifier, const char * name, size_t do
   }
   memcpy(const_cast<char *>(node_handle->name), name, strlen(name) + 1);
 
+  node_handle->name_space =
+    reinterpret_cast<const char *>(rmw_allocate(sizeof(char) * strlen(name_space) + 1));
+  if (!node_handle->name_space) {
+    RMW_SET_ERROR_MSG("failed to allocate memory for node name");
+    goto fail;
+  }
+  memcpy(const_cast<char *>(node_handle->name_space), name_space, strlen(name_space) + 1);
+
   buf = rmw_allocate(sizeof(ConnextNodeInfo));
   if (!buf) {
     RMW_SET_ERROR_MSG("failed to allocate memory");
@@ -475,6 +491,9 @@ fail:
   if (node_handle) {
     if (node_handle->name) {
       rmw_free(const_cast<char *>(node_handle->name));
+    }
+    if (node_handle->name_space) {
+      rmw_free(const_cast<char *>(node_handle->name_space));
     }
     rmw_free(node_handle);
   }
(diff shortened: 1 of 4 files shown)


rmw_connext_shared_cpp/src/shared_functions.cpp
```diff
@@ -430,6 +438,14 @@ create_node(const char * implementation_identifier, const char * name, size_t do
   }
   memcpy(const_cast<char *>(node_handle->name), name, strlen(name) + 1);
 
+  node_handle->name_space =
+    reinterpret_cast<const char *>(rmw_allocate(sizeof(char) * strlen(name_space) + 1));
+  if (!node_handle->name_space) {
+    RMW_SET_ERROR_MSG("failed to allocate memory for node name");
```
dirk-thomas: Copy-n-paste error?

wjwwood: Thanks fixed in 930df8f.

rmw_connext_shared_cpp/src/shared_functions.cpp
```diff
@@ -319,6 +323,10 @@ create_node(const char * implementation_identifier, const char * name, size_t do
   }
   participant_qos.participant_name.name =
     static_cast<char *>(rmw_allocate((strlen(name) + 1) * sizeof(char)));
+  if (!participant_qos.participant_name.name) {
+    RMW_SET_ERROR_MSG("failed to allocate memory");
+    return NULL;
+  }
   participant_qos.participant_name.name = strdup(name);
```
Karsten1987: isn't that a potential memory leak here? We call rmw_allocate beforehand and then in line 330 `strdup` allocates again memory for the string internally. 
Sorry for comment on this line, it's technically not part of the PR.

rmw_connext_shared_cpp/src/shared_functions.cpp
```diff
@@ -430,6 +438,14 @@ create_node(const char * implementation_identifier, const char * name, size_t do
   }
   memcpy(const_cast<char *>(node_handle->name), name, strlen(name) + 1);
 
+  node_handle->namespace_ =
+    reinterpret_cast<const char *>(rmw_allocate(sizeof(char) * strlen(namespace_) + 1));
+  if (!node_handle->namespace_) {
+    RMW_SET_ERROR_MSG("failed to allocate memory for node namespace");
+    goto fail;
+  }
+  memcpy(const_cast<char *>(node_handle->namespace_), namespace_, strlen(namespace_) + 1);
```
Karsten1987: this is link to the previous comment, where I believe that `memcpy` is the right approach here and `strdup` isn't.

rmw_connext_shared_cpp/src/shared_functions.cpp
```diff
@@ -555,6 +574,8 @@ destroy_node(const char * implementation_identifier, rmw_node_t * node)
   node->data = nullptr;
   rmw_free(const_cast<char *>(node->name));
   node->name = nullptr;
+  rmw_free(const_cast<char *>(node->namespace_));
```
Karsten1987: didn't look up the complete `node` struct here, but what's the difference between the `node` and the `node_handle` is this context?

rmw_connext_shared_cpp/src/shared_functions.cpp
wjwwood: Yes... I was just adding the check that I thought was missing. But the malloc looks completely unnecessary and leaked because we use `strdup` there.

wjwwood: Oh dear:

https://github.com/ros2/rmw_connext/blame/master/rmw_connext_shared_cpp/src/shared_functions.cpp#L322

😛

wjwwood: I'll fix it in this branch.

rmw_connext_shared_cpp/src/shared_functions.cpp
wjwwood: `node_handle` is just the variable name in the `create_node` function. In this function, `destroy_node` that same object is passed in as `node` rather than `node_handle`. At least that's how I read the code.

rmw_connext_shared_cpp/src/shared_functions.cpp
wjwwood: See 27ffa13 for my guess at a fix, I'm still waiting for it to rebuild so I can check for double free's.

Karsten1987: man, git blame, that beast :laughing:

wjwwood: I had to remove the extra free. DDS appears to free this for us, because I get this on exit if I leave the free in:

```
[D0087|DELETE Sub]RTIOsapiHeap_freeBufferAligned:inconsistent free/alloc: block id 0X1 being freed with "RTIOsapiHeap_freeBufferAligned" and was allocated with "RTIOsapiHeap_unknownFunction"
talker(63898,0x7fff72c12000) malloc: *** error for object 0x7fc288c17cc0: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
```

See: 44fb031

PR #221, opened 2017-04-01
Namespace impl

Karsten1987:
connects to ros2/ros2#327


rmw_connext_cpp/src/functions.cpp
```diff
@@ -999,15 +1195,14 @@ rmw_create_client(
   RMW_TRY_PLACEMENT_NEW(client_info, buf, goto fail, ConnextStaticClientInfo, )
   buf = nullptr;  // Only free the client_info pointer; don't need the buf pointer anymore.
   client_info->requester_ = requester;
-  client_info->callbacks_ = callbacks;
-  client_info->response_datareader_ = response_datareader;
+  client_info->callbacks_ = callbacks;  client_info->response_datareader_ = response_datareader;
```
Karsten1987: cleanup

rmw_connext_cpp/src/functions.cpp
```diff
@@ -695,6 +794,15 @@ rmw_create_subscription(const rmw_node_t * node,
   if (buf) {
     rmw_free(buf);
   }
+
+  // cleanup namespacing
+  if (utilities_string_array_fini(&name_tokens) != UTILITIES_RET_OK) {
+    fprintf(stderr, "Failed to destroy the token string array\n");
+  }
+  // if (partition_str) {
```
Karsten1987: cleanup

rmw_connext_cpp/src/functions.cpp
```diff
@@ -577,13 +641,47 @@ rmw_create_subscription(const rmw_node_t * node,
...
+
+  dds_subscriber = participant->create_subscriber(
+    subscriber_qos, NULL, DDS_STATUS_MASK_NONE);
   if (!dds_subscriber) {
     RMW_SET_ERROR_MSG("failed to create subscriber");
     goto fail;
   }
 
-  topic_description = participant->lookup_topicdescription(topic_name);
+  // TODO(karsten1987): I have to verify how this topic gets found
```
Karsten1987: remove comments. works.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -971,9 +1090,54 @@ rmw_create_client(
...
+    rmw_free(response_concat_str);
+  } else {
+    RMW_SET_ERROR_MSG("Illformated service name");
+    goto fail;
+  }
+  // all necessary strings are copied into connext
+  // free that memory
+  utilities_string_array_fini(&name_tokens);
+
+  // TODO(karsten1987): For now, I'll expose the datawriter
```
Karsten1987: remove comment. works.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -1050,6 +1243,18 @@ rmw_create_client(
   if (buf) {
     rmw_free(buf);
   }
+
+  // cleanup namespacing
+  if (utilities_string_array_fini(&name_tokens) != UTILITIES_RET_OK) {
+    fprintf(stderr, "Failed to destroy the token string array\n");
+  }
+  // if(request_partition_str) {
```
Karsten1987: cleanup

rmw_connext_cpp/src/functions.cpp
```diff
@@ -1316,6 +1604,18 @@ rmw_create_service(
...
```
Karsten1987: cleanup

rmw_connext_cpp/CMakeLists.txt
```diff
@@ -56,7 +59,9 @@ register_rmw_implementation(
 )
 
 add_library(rmw_connext_cpp SHARED src/functions.cpp)
+target_link_libraries(rmw_connext_cpp ${rcutils_LIBRARIES})
```
dirk-thomas: This line is redundant with `rcutils` being listed two lines below with `ament_target_dependencies`.

rmw_connext_cpp/CMakeLists.txt
```diff
@@ -81,6 +86,25 @@ endif()
...
+  #  #foreach(typesupport_impl_cpp ${typesupport_impls_cpp})
+  #  #  rosidl_target_interfaces(test_externally_defined_services
+  #  #    mock_msgs ${typesupport_impl_cpp})
+  #  #endforeach()
+  #  #foreach(typesupport_impl_c ${typesupport_impls_c})
+  #  #  rosidl_target_interfaces(test_externally_defined_services
+  #  #    mock_msgs ${typesupport_impl_c})
+  #  #endforeach()
+  #endif()
+
```
dirk-thomas: Are you planning on enabling this before merging? If not should it either be removed or a TODO added above?

rmw_connext_cpp/package.xml
```diff
@@ -14,6 +14,7 @@
   <buildtool_export_depend>rosidl_cmake</buildtool_export_depend>
 
   <build_depend>connext_cmake_module</build_depend>
+  <build_depend>rcutils</build_depend>
```
dirk-thomas: Alpha order.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -146,8 +150,11 @@ to_string(DDS_InstanceHandle_t val)
 
 extern "C"
 {
-const char * rti_connext_identifier = "rmw_connext_cpp";
-
+// static for internal linkage
+static const char * const rti_connext_identifier = "rmw_connext_cpp";
+static const char * const ros_topics_prefix = "rt";
+static const char * const ros_service_requester_prefix = "rqtacco";
```
dirk-thomas: Why is the value `rqtacco` (sorry if I missed that in the design discussion)?

rmw_connext_cpp/src/functions.cpp
```diff
@@ -225,9 +232,13 @@ rmw_create_publisher(
 
   RMW_CONNEXT_EXTRACT_MESSAGE_TYPESUPPORT(type_supports, type_support)
 
+  if (!topic_name || strlen(topic_name) == 0) {
+    RMW_SET_ERROR_MSG("publisher topic is null or empty string");
```
dirk-thomas: Should this `return NULL`?

rmw_connext_cpp/src/functions.cpp
```diff
@@ -361,6 +410,14 @@ rmw_create_publisher(
     dds_publisher->get_instance_handle(), topic_name, type_name, EntityType::Publisher);
   node_info->publisher_listener->trigger_graph_guard_condition();
 
+#ifdef DISCOVERY_DEBUG_LOGGING
+  fprintf(stderr, "******* Creating Publisher Details: ********\n");
+  fprintf(stderr, "Publisher partition %s\n", publisher_qos.partition.name[0]);
+  fprintf(stderr, "Publisher topic %s\n", topic_writer->get_topic()->get_name());
+  fprintf(stderr, "Publisher address %p\n", static_cast<void *>(dds_publisher));
+  fprintf(stderr, "******\n");
+#endif
```
dirk-thomas: Can you add a TODO for this block to replace this with logging macros in the future.

rmw_connext_cpp/test/test_namespace_publisher.cpp
```diff
@@ -0,0 +1,27 @@
...
+class TestNamespacePublisher : public ::testing::Test
+{
+protected:
+  static void SetUpTestCase()
+  {
+  }
+};
+
+TEST_F(TestNamespacePublisher, empty_namespace) {
+  FAIL();
```
dirk-thomas: Should this be filled with an actual test?

rmw_connext_cpp/src/functions.cpp
wjwwood: I'd say there is no need, because when the logging macros are available we'll have to audit all of the code for uses of printf and fprintf to see if they should be replaced.

rmw_connext_cpp/package.xml
wjwwood: fa82cb5

wjwwood:
I'm not going to re-review all of this since @dirk-thomas has already started and has outstanding comments.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -279,14 +297,45 @@ rmw_create_publisher(
...
+    partition_str = DDS_String_dup(ros_topics_prefix);
+    topic_str = DDS_String_dup(name_tokens.data[0]);
+  } else if (name_tokens.size == 2) {
+    // TODO(Karsten1987): Fix utility function for this
+    size_t partition_length = strlen(ros_topics_prefix) + strlen(name_tokens.data[0]) + 2;
+    char * concat_str = reinterpret_cast<char *>(rmw_allocate(partition_length * sizeof(char)));
+    snprintf(concat_str, partition_length, "%s/%s", ros_topics_prefix, name_tokens.data[0]);
+    // Connext will call deallocate on this, passing ownership to connext
+    partition_str = DDS_String_dup(concat_str);
+    topic_str = DDS_String_dup(name_tokens.data[1]);
```
dirk-thomas: See https://github.com/ros2/rmw_connext/pull/227#discussion_r136463024 It looks like that the variable is not freed and `valgrind` marks it as a leak. Any idea why this was added / for which case?

wjwwood: It could be that `partition_str` is actually taking ownership, but `topic_str` is not and needs to be explicitly free'd.

I remember debugging this with @Karsten1987 and he found that Connext was taking ownership of some string, though I don't remember which or both.

Karsten1987: as far as i remember, the reason why we call DDS_String_dup on it is because connnext takes ownership. We duplicate the string so that we can delete the initial allocated memory (`concat_str`, `name_tokens`)

dirk-thomas: See #244.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -1829,6 +2180,7 @@ rmw_service_server_is_available(
     number_of_response_publishers);
 #endif
   if (number_of_response_publishers == 0) {
+    fprintf(stderr, "Number of response publishers is 0\n");
```
wjwwood: This gets printed in normal use cases... https://github.com/ros2/rmw_connext/pull/322

PR #222, opened 2017-04-11
update mapping of char to signed char

dirk-thomas:
Fixes the current compile error of `rmw_connext_dynamic_cpp` introduced by ros2/rosidl#211. The rmw impl. still doesn't "work" but at least it is compilable again.

[Build Status](http://ci.ros2.org/job/ci_linux/2434/)


dirk-thomas:
I addressed the style problems. The failing tests are expected.

gerkey:
Whoops, sorry for breaking that. Was there a way for me to know that I'd broken it?

mikaelarguedas:
We can enable/disable arbitrary rmw implementation on CI using checkboxes. But we don't test connext_dynamic and opensplice by default on CI so you couldn't know. Most changes don't affect non-supported rmw_implementation. The type of changes affecting all rmw_implementations:
- changes in rmw
- changes in idl/types (anything related to type support)

Very likely many others but I'd say these are the main ones I faced in the past

Karsten1987:
I noticed that by compiling my workspace locally. I know that tests don't
run for dynamic (hence the disable button on the farm), but the workspace
should still be compiling.

PR #225, opened 2017-04-18
DDS Security support

mikaelarguedas:
connects to ros2/ros2#339

### DIFF
--- a/rmw_connext_dynamic_cpp/src/functions.cpp
+++ b/rmw_connext_dynamic_cpp/src/functions.cpp
@@ -204,9 +204,12 @@ rmw_create_node(
   const char * name,
   const char * namespace_,
   size_t domain_id,
+  bool enforce_security,
   const char * security_root_path)
 {
-  return create_node(rti_connext_dynamic_identifier, name, namespace_, domain_id, security_root_path);
+  return create_node(
+    rti_connext_dynamic_identifier, name, namespace_, domain_id,
+    enforce_security, security_root_path);
 }
 
 rmw_ret_t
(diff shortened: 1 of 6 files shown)


rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -198,9 +198,20 @@ rmw_init()
...
 
 rmw_node_t *
+rmw_create_secure_node(
+  const char * name,
+  const char * namespace_,
+  size_t domain_id,
+  const char * security_files_paths)
+{
+  (void)security_files_paths;
+  return create_node(rti_connext_dynamic_identifier, name, namespace_, domain_id, NULL);
```
dirk-thomas: Shouldn't this return an error instead?

mikaelarguedas: given that connext_dynamic is broken I don't know if this works or not so I didn't pass the variable. I'll pass it anyway given that the code path for connext and connext dynamic is exactly the same in the case of node creation. Thanks for pointing it out

rmw_connext_shared_cpp/src/shared_functions.cpp
```diff
@@ -346,6 +347,103 @@ create_node(
     return NULL;
   }
 
+  if (security_options->security_root_path) {
```
dirk-thomas: You must check that `security_options` is not null before accessing its members.

mikaelarguedas: done in https://github.com/ros2/rmw_connext/pull/225/commits/dda0cd0b30fa03eb08e5e23ef4d7d6b5a4560198

PR #227, opened 2017-05-31
implement avoid_ros_namespace_conventions QoS option

wjwwood:
This implements the `avoid_ros_namespace_conventions` QoS option. @Karsten1987 I also refactored this to share code between the pub-sub and again between service-client. It could be deduplicated more, but I think this is incrementally better. I also switched to using `rcutils_format_string` to address one of your TODO's. Please have a look at those changes carefully when reviewing (lots of copy-paste going on, could be a mistake).

Connects to ros2/rmw#106

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -40,8 +40,7 @@
 # pragma GCC diagnostic pop
 #endif
 
-// TODO(karsten1987): Introduce rcutils.h
-#include "rcutils/concat.h"
+#include "rcutils/format_string.h"
 #include "rcutils/types.h"
 #include "rcutils/split.h"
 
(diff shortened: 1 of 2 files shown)


rmw_connext_cpp/src/functions.cpp
```diff
@@ -214,6 +213,57 @@ rmw_destroy_node(rmw_node_t * node)
   return destroy_node(rti_connext_identifier, node);
 }
 
+bool
+__process_topic_name(
```
Karsten1987: nitpick, but I think this file uses only one underscore for "non-public" methods.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -214,6 +213,57 @@ rmw_destroy_node(rmw_node_t * node)
...
+        success = false;
+        goto end;
+      }
+      *partition_str = DDS_String_dup(concat_str);
+      allocator.deallocate(concat_str, allocator.state);
+    }
+    // Connext will call deallocate on this, passing ownership to connext
+    *topic_str = DDS_String_dup(name_tokens.data[1]);
+  } else {
+    RMW_SET_ERROR_MSG("Illformated topic name")
```
Karsten1987: That might be my incorrect english :)

rmw_connext_cpp/src/functions.cpp
```diff
@@ -214,6 +213,57 @@ rmw_destroy_node(rmw_node_t * node)
...
 
+bool
+__process_topic_name(
+  const char * topic_name,
+  bool avoid_ros_namespace_conventions,
+  char ** topic_str,
+  char ** partition_str)
+{
+  bool success = true;
+  rcutils_string_array_t name_tokens;
```
Karsten1987: I should have default initialized that data structure as done in the service partitions.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -1023,6 +1037,71 @@ rmw_wait(rmw_subscriptions_t * subscriptions,
...
+      if (!request_concat_str) {
+        RMW_SET_ERROR_MSG("could not allocate memory for partition string")
+        success = false;
+        goto end;
+      }
+      char * response_concat_str = rcutils_format_string(
+        allocator,
+        "%s/%s", ros_service_response_prefix, name_tokens.data[0]);
+      if (!response_concat_str) {
+        allocator.deallocate(request_concat_str, allocator.state);
```
wjwwood: No, the `request_concat_str` points to allocated memory at this point. If the deallocate did not happen, then it would be leaked here. There is nothing that needs to be deallocated in either of the other two cases, `request_concat_str` or `concat_str`.

dhood: yep, thanks. I read it as deallocating `response_concat_str`.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -297,36 +346,21 @@ rmw_create_publisher(
...
-    partition_str = DDS_String_dup(concat_str);
-    topic_str = DDS_String_dup(name_tokens.data[1]);
-    // free temporary memory
-    rmw_free(concat_str);
-  } else {
-    RMW_SET_ERROR_MSG("Illformated topic name");
+  if (!_process_topic_name(
+      topic_name,
+      qos_profile->avoid_ros_namespace_conventions,
+      &topic_str,
```
dirk-thomas: @wjwwood When is the allocated memory of this variable being freed? Same for `partition_str` next line.

rmw_connext_cpp/src/functions.cpp
```diff
@@ -649,36 +678,21 @@ rmw_create_subscription(const rmw_node_t * node,
...
```
dirk-thomas: @wjwwood When is the allocated memory of this variable being freed? Same for `partition_str` next line.

rmw_connext_cpp/src/functions.cpp
wjwwood: Connext takes ownership of it according to this comment: https://github.com/ros2/rmw_connext/commit/69bc3e61b919ce5e88348456473917e12ead86dc#diff-becdf01440f1757df19f4775b534e066R251

I believe if you free it then connext later does and you get a double free.

dirk-thomas: Currently it shows up as being leaked in `valgrind` with the `demo_nodes_cpp` `talker` / `listener`.

wjwwood: You can try freeing it, but I'm pretty sure we were getting a double free before.

wjwwood: Also, this is the wrong pull request to ask on I think because it was only moved into a function in this pr. The removed code already had this ownership comment in place.

mikaelarguedas: for reference, seems related to https://github.com/ros2/rcl/pull/161/files, maybe doing one will impact the other

dirk-thomas: > for reference, seems related to https://github.com/ros2/rcl/pull/161/files, maybe doing one will impact the other

Those are not related. `topic_str` is a local variable within `rmw_connext_cpp`.

dirk-thomas: Moved: https://github.com/ros2/rmw_connext/pull/221#discussion_r136465335

PR #228, opened 2017-05-31
Topic names

Karsten1987:
connects to ros2/rmw_fastrtps#114

this should fix the inconsistency with `rclcpp::Node::count_subscribers` and `rclcpp::Node::count_publishers` as well as `get_topic_names_and_types`.

The function truncates the internal topic fqdn in the form of `rt/my_namespace/my_topic` into `/my_namespace/my_topic`

### DIFF
--- a/rmw_connext_cpp/src/functions.cpp
+++ b/rmw_connext_cpp/src/functions.cpp
@@ -1878,7 +1878,10 @@ rmw_get_topic_names_and_types(
   rmw_topic_names_and_types_t * topic_names_and_types)
 {
   return get_topic_names_and_types(rti_connext_identifier, node,
-           topic_names_and_types);
+           topic_names_and_types,
+           ros_topics_prefix,
+           ros_service_requester_prefix,
+           ros_service_response_prefix);
 }
 
 rmw_ret_t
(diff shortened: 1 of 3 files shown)


rmw_connext_cpp/src/functions.cpp
```diff
@@ -1878,7 +1878,10 @@ rmw_get_topic_names_and_types(
   rmw_topic_names_and_types_t * topic_names_and_types)
 {
   return get_topic_names_and_types(rti_connext_identifier, node,
-           topic_names_and_types);
+           topic_names_and_types,
+           ros_topics_prefix,
+           ros_service_requester_prefix,
+           ros_service_response_prefix);
```
mikaelarguedas: please wrap after the open parenthesis and use 2 spaces indentation (same everywhere below)

Karsten1987: I left it like this on purpose for not blowing up the diff. The style format here should be excluded in a different PR IMO.

mikaelarguedas: Fair enough.
Would you mind doing it for the cases where the functions were fitting in one line before this PR?

Karsten1987: I'll open a new PR for style on this.

mikaelarguedas: :+1:

wjwwood:
Since this is reviewed and tested, and since the connected pr's have already been merged I'm going to merge this for you @Karsten1987.

dirk-thomas:
As of this PR `rmw_connext_dynamic_cpp` fails to build.

PR #229, opened 2017-06-07
use split allocator

Karsten1987:
connects to ros2/rcutils#6


dirk-thomas:
This needs to be rebased.

wjwwood:
I rebased it, since it was undercut by my other changes.

wjwwood:
@mikaelarguedas or @dirk-thomas can you re-review this since I rebased?

Issue #236, opened 2017-07-12
refactor functions.cpp


Karsten1987:
Refactor rmw_connext_cpp's `functions.cpp` into multiple files.

PR #237, opened 2017-07-12
Refactor functions cpp

Karsten1987:
connects to ros2/rmw_connext#236

1.) split `functions.cpp` into multiple files (preserving history)
2.) cleanup new files with their respective parts
3.) change `CMakeLists.txt` to compile new files

4.) split `shared_functions.hpp` into multiple header files (keeping original file as a general include)
5.) cleanup new header files with their respective parts

6.) rework connext_cpp to use only individual header files
7.) apply style changes for functions


Karsten1987:
linux: [Build Status](http://ci.ros2.org/job/ci_linux/2905/) (unrelated test failures)
osx: [Build Status](http://ci.ros2.org/job/ci_osx/2344/) (unrelated test failures)
windows: [Build Status](http://ci.ros2.org/job/ci_windows/3028/)

Karsten1987:
I think this is up for review. I tried to keep the commits as separate as possible. Not sure what we should squash here.

dirk-thomas:
Are you sure the failing `rclpytests` are unrelated? What are you basing that assumption on? They don't fail with the nightly builds and seem to "consistently" (two out of two finished jobs) fail with this PR.

Karsten1987:
I based it on the fact that the windows build was green. But fair point, I'll have a look into it.

Karsten1987:
Not sure, why the python tests are failing, but i'll have a look into it.

Karsten1987:
So, looking into the latest nighly builds, these tests also fail there:
http://ci.ros2.org/view/nightly/job/nightly_linux_repeated/lastCompletedBuild/consoleFull
```
10:02:24 1: .........Exception ignored in: <bound method Node.__del__ of <rclpy.node.Node object at 0x7f849a6eeda0>>```

I just compiled my workspace with the current master of connext and rclpy also fails.

**Comment by dirk-thomas** [comment]
> So, looking into the latest nighly builds, these tests also fail there:

The console output of the nightly shows the exception but in that build the `rclpytests` did pass.

@Karsten1987 and I looked into this and came up with ros2/rclpy#99 to get rid of this exception by fixing the logic in one of the `rclpy` tests.

**Comment by dirk-thomas** [comment]
linux: [Build Status](http://ci.ros2.org/job/ci_linux/2907/)
osx: [Build Status](http://ci.ros2.org/job/ci_osx/2350/)

**Comment by mikaelarguedas** [comment]
The linux aarch64 build failed last time this PR ran ci so I triggered one arm 64 job to see if this is still the case
linux-aarch64: [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/387/)

**Comment by dirk-thomas** [comment]
@mikaelarguedas We can't run Connext on ARM therefore there is no RMW impl. available and the build fails.

**Comment by mikaelarguedas** [comment]
Yep just realized that with @Karsten1987 (see previous comment)

**Comment by dirk-thomas** [comment]
After increasing the timeout for the `rclpy` tests (ros2/rclpy#100):

linux: [Build Status](http://ci.ros2.org/job/ci_linux/2908/)
osx: [Build Status](http://ci.ros2.org/job/ci_osx/2351/)

File: rmw_connext_cpp/src/rmw_client.cpp [hunk 1]
```diff
@@ -0,0 +1,338 @@
...
+#include "rmw/error_handling.h"
+#include "rmw/impl/cpp/macros.hpp"
+#include "rmw/rmw.h"
+
+#include "rmw_connext_shared_cpp/types.hpp"
+#include "rmw_connext_shared_cpp/qos.hpp"
+
+#include "identifier.hpp"
+#include "process_topic_and_service_names.hpp"
+#include "type_support_commons.hpp"
```
dhood: apologies for being late to the party, but are the any objections if we change "commons" to "common"? I thought it was a typo the first time I saw it. I can make a followup PRs

Karsten1987: this is fine with me, if you have a completely different name for the file which describes it better, feel free :)

Karsten1987:
connext_cpp doesn't have any public headers, which other packages could/should include. Now, with that refactoring, there is commonly used code between multiple cpp files which needs a header file. Does that make sense?

dhood:
thanks for the explanation, that makes perfect sense!

rmw_connext_cpp/src/rmw_client.cpp
dhood: thanks! https://github.com/ros2/rmw_connext/pull/240 (it was just that commons has a different meaning)

mikaelarguedas:
@Karsten1987 Is the refactor of `connext_dynamic_cpp` planned in a follow-up PR ?

Karsten1987:
we should, but I don't think I'll work on it in the near future. I'll open an issue for visibility.

mikaelarguedas:
sounds good, hopefully it will allow us to refactor more things into the rmw_connext_shared_cpp. Thanks!

Karsten1987:
ros2/rmw_connext#241

PR #242, opened 2017-08-18
Update the Connext license filter resource name to specify it's a prefix

dhood:
connects to ros2/launch#59


rmw_connext_shared_cpp/cmake/get_rmw_connext_output_filter.cmake
```diff
@@ -15,8 +15,9 @@
 #
 # Get the output filter for the Connext RMW packages.
 #
-# :param output_filter: the filter which will be used to identify console
-#   output generated by Connext
+# :param output_filter: the filter that will be used to identify console
+#   output generated by Connext. This is provided as a single string with each
+#   line of the string containing a filter that will be checked indepdently.
```
dirk-thomas: Spelling: indepdently.

dhood:
the motivation of the attached pr was to be able to filter things like the fastrtps warning that would require a regex. therefore I think it's appropriate for internal filters to be regexes not prefixes if we are going to choose one to support

dirk-thomas:
> I think it's appropriate for internal filters to be regexes not prefixes if we are going to choose one to support

Why not leave that decision to the rmw implementation. They can choose if they want to use a prefix filter or a regex filter - depending on what is necessary / sufficient.

PR #243, opened 2017-08-21
Expose native handles

dirk-thomas:
Similar to ros2/rmw_fastrtps#145 and ros2/rmw_fastrtps#146.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3059/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2458/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3130/)

### DIFF
--- a/rmw_connext_cpp/src/identifier.cpp
+++ b/rmw_connext_cpp/src/identifier.cpp
@@ -12,9 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "identifier.hpp"
+#include "rmw_connext_cpp/identifier.hpp"
 
-extern "C"
-{
 const char * const rti_connext_identifier = "rmw_connext_cpp";
-}  // extern "C"
--- a/rmw_connext_cpp/src/rmw_client.cpp
+++ b/rmw_connext_cpp/src/rmw_client.cpp
@@ -22,10 +22,10 @@
 #include "rmw_connext_shared_cpp/types.hpp"
 #include "rmw_connext_shared_cpp/qos.hpp"
 
-#include "identifier.hpp"
+#include "rmw_connext_cpp/connext_static_client_info.hpp"
+#include "rmw_connext_cpp/identifier.hpp"
 #include "process_topic_and_service_names.hpp"
 #include "type_support_common.hpp"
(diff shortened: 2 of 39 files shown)


rmw_connext_cpp/src/identifier.cpp
```diff
@@ -12,9 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "identifier.hpp"
+#include "rmw_connext_cpp/identifier.hpp"
 
-extern "C"
-{
```
mikaelarguedas: does it hurt to keep it ? otherwise if we ever add functions here we will have to remember to add back the extern

rmw_connext_cpp/src/rmw_client.cpp
```diff
@@ -22,10 +22,10 @@
 #include "rmw_connext_shared_cpp/types.hpp"
 #include "rmw_connext_shared_cpp/qos.hpp"
 
-#include "identifier.hpp"
+#include "rmw_connext_cpp/identifier.hpp"
 #include "process_topic_and_service_names.hpp"
 #include "type_support_common.hpp"
-#include "types/connext_static_client_info.hpp"
+#include "rmw_connext_cpp/connext_static_client_info.hpp"
```
mikaelarguedas: Nit: include order (same in other files below)

rmw_connext_cpp/src/identifier.cpp
dirk-thomas: I don't see a reason to declare the symbol `extern "C"`. Also none of the other RMW impl. does that.

rmw_connext_cpp/src/rmw_client.cpp
dirk-thomas: I fixed the order of this include but as far as I see there are many more includes not following that rule since it is not enforced by the linter atm.

mikaelarguedas: Agreed, I made the comment because your analog PR in rmw_fastrtps was complying and not this one. Apparently you fixed only this one though and not the other ones in this PR

PR #250, opened 2017-09-14
MacOS: look for libs in newer subfolders

dirk-thomas:
Removed `x64Darwin14clang6.0` since it wasn't used for a long time already (Connext < 5.2).

Added `x64Darwin16clang8.0` since it is used for Connext 5.3.0.


dhood:
darwin14 + clang6.0 is supported for 5.2.3 (inferred from http://s3.amazonaws.com/RTI/Bundles/5.2.3/Evaluation/rti_connext_dds-5.2.3-eval-x64Darwin14clang6.0.dmg being a valid download). Are we looking to drop support for 5.2.3?

dirk-thomas:
While it is possible to use that version all our Jenkins nodes use `Darwin15clang7` so I think it is safe to remove the previous one. We will likely only support 5.2.3 for a very short time since the pending security PRs will likely be merged soon which will make us require 5.3.0.

dhood:
OK; tbh I have been using darwin15 connext on darwin14 without any issues anyway. if that changes for 5.3.0 it's time for me to update :)

PR #253, opened 2017-10-02
add argument to invocation to get returncode 0 for both connext 5.2.4…

mikaelarguedas:
… and 5.3.0

I was looking into why my local builds became so much slower since I switched to connext 5.3.0.
The part that seems much slower is the invocation of the `rtiddsgen_server` executable in the Connext cmake module.

While I haven't found yet how to improve the speed in the invocation of the generator (they seem to be just as fast when running them in a shell), I noticed that none of our builds were using the `ddsgen_server` since we upgraded. The reason is that the return code when invoking the executable without argument is not the same between 5.2.X and 5.3.0.
Adding the `-version` argument seems to bring us back to both returning the code 0.

You can confirm it by looking for "/bin/rtiddsgen" in:
the last nightly: http://ci.ros2.org/view/nightly/job/nightly_linux_debug/602/consoleFull
the job with this change: http://ci.ros2.org/job/ci_linux/3295/consoleFull

### DIFF
--- a/connext_cmake_module/cmake/Modules/FindConnext.cmake
+++ b/connext_cmake_module/cmake/Modules/FindConnext.cmake
@@ -327,7 +327,7 @@ endif()
 if(Connext_DDSGEN_SERVER)
   # check that the generator is invocable / finds a Java runtime environment
   execute_process(
-    COMMAND "${Connext_DDSGEN_SERVER}"
+    COMMAND "${Connext_DDSGEN_SERVER}" "-version"
     RESULT_VARIABLE _retcode
     OUTPUT_QUIET ERROR_QUIET)
   if(NOT _retcode EQUAL 0)


mikaelarguedas:
Ok looking a bit more into this it looks like the executable `rtiddsgen_server` behaves differently between 5.2.X and 5.3.0.

5.2.X it fails earlier if not passed any argument:
5.2.3:
- no arguments:
```
real	0m0.197s
user	0m0.008s
sys	0m0.004s
```
- with `-version`:
```
real	0m0.705s
user	0m0.004s
sys	0m0.008s
```

For connext 5.3.0 it seems to be roughly the same time:
 - no arguments:
```
real	0m0.525s
user	0m0.000s
sys	0m0.012s
```
  - with `-version`:
```
real	0m0.453s
user	0m0.004s
sys	0m0.008s
```

When running the same commands through CMake `execute_process()`:
- 5.2.3:
  - no arguments: <1sec
  - with `-version`: ~20sec
- 5.3.0:
  - no arguments: ~20sec
  - with `-version`: ~20sec

So that's where the slow down in the CMake configure step comes from. Not sure what we can do in CMake to speed that up. I think this PR should be merged as is to allow the use of `rtiddsgen_server` on our current CIs. If I find a workaround to reduce the execution time I'll open a follow-up PR

mikaelarguedas:
Note to self: We don't use `rtiddsgen_server` on MacOS because the executable doesnt seem to be there in 5.3.0 (but is present in 5.2.3)

mikaelarguedas:
Final note on this: RTI confirmed that they don't support or ship `rtiddsgen_server` for Mac as of 5.3.0. There was a hiccup and it did not make it to the release notes

PR #255, opened 2017-10-06
fix hang of server waiting for forked daemon to time out

### LINKED ISSUE #254: FindConnext much slower since 5.3.0 upgrade
Finding Connext was taking a few seconds and is now taking ~30sec since we upgraded to Connext 5.3.0.
The main reason seems to be the execution time of `rtiddsgen_server` invoked from CMake.

See https://github.com/ros2/rmw_connext/pull/253#issuecomment-333705170 for more details

dirk-thomas:
Temporarily fixes #254.

### DIFF
--- a/connext_cmake_module/cmake/Modules/FindConnext.cmake
+++ b/connext_cmake_module/cmake/Modules/FindConnext.cmake
@@ -325,6 +325,12 @@ if(Connext_FOUND)
 endif()
 
 if(Connext_DDSGEN_SERVER)
+  # the first invocation with a timeout is necessary since when the script
+  # forks the daemon it fails to exit early (default daemon timeout is 20s)
+  execute_process(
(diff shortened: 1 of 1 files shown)


dirk-thomas:
> Maybe we should add a TODO or open another ticket to remind us to remove the workaround in the future

I don't think this needs a todo in the code. Whenever we get an updated version of Connext which addresses the problem we can remove the patch.

mikaelarguedas:
In several places we track upstream issues otherwise we never revert the hacks, I'm not sure why this particular one is different

dirk-thomas:
Please feel free to add a todo if you think it is necessary.

mikaelarguedas:
Sounds good https://github.com/ros2/rmw_connext/commit/2edfe9da0ced9b9e44382a773216fb01e2768b4a

PR #256, opened 2017-10-09
use undocumented option to avoid need for timeout

dirk-thomas:
Follow up of #255.


mikaelarguedas:
tested locally on linux it does fix the issue, taking less than a second for each invocation :tada: .
Do you want to run a job on the other platforms to make sure it works across the board before merging ?

dirk-thomas:
> Do you want to run a job on the other platforms to make sure it works across the board before merging ?

Since that is essentially Java code I didn't plan to run CI for the other platforms.

mikaelarguedas:
> Since that is essentially Java code I didn't plan to run CI for the other platforms.
 
Sounds good to me :+1:

PR #258, opened 2017-10-23
Remove rmw_connext dead code

gaoethan:
Dead code in API `rmw_create_subscription()`
the condition "`subscription`" cannot be true in the "`fail`"
clauses and the execution never reaches this statement:

```
if (subscription) {
    rmw_subscription_free(subscription);
    }
```

the "`subscribtion`" will be finally free by
`rmw_destroy_subscription()` when it's not a nullptr

Dead code in API `rmw_create_publisher()`
the condition "`publisher`" cannot be true in the "`fail`"
clauses, so the execution cannot reach this statement:

```
if (publisher) {
    rmw_publisher_free(publisher);
    }
```

the "`publisher`" will be finally free by
`rmw_destroy_publisher()` when it's not a nullptr

Signed-off-by: Ethan Gao <ethan.gao@linux.intel.com>

### DIFF
--- a/rmw_connext_cpp/src/rmw_publisher.cpp
+++ b/rmw_connext_cpp/src/rmw_publisher.cpp
@@ -252,9 +252,7 @@ fail:
     DDS_String_free(topic_str);
     topic_str = nullptr;
   }
-  if (publisher) {
-    rmw_publisher_free(publisher);
-  }
+
   // Assumption: participant is valid.
   if (dds_publisher) {
     if (topic_writer) {
--- a/rmw_connext_cpp/src/rmw_subscription.cpp
+++ b/rmw_connext_cpp/src/rmw_subscription.cpp
@@ -250,9 +250,7 @@ fail:
     DDS_String_free(topic_str);
     topic_str = nullptr;
   }
-  if (subscription) {
-    rmw_subscription_free(subscription);
-  }
+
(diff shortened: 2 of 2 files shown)


Karsten1987:
I may miss something here, but I think this PR is not correct. Opposite to your PR in `rmw_fastrtps_cpp` https://github.com/ros2/rmw_fastrtps/pull/163, in `rmw_connext_cpp` the memory allocation for the publisher happens first [0] and thus still can end up in a `fail` state.

[0] https://github.com/ros2/rmw_connext/blob/master/rmw_connext_cpp/src/rmw_publisher.cpp#L102

rmw_connext_cpp/src/rmw_publisher.cpp
```diff
@@ -252,9 +252,7 @@ rmw_create_publisher(
     DDS_String_free(topic_str);
     topic_str = nullptr;
   }
-  if (publisher) {
```
dirk-thomas: This is necessary of the goto in this line is triggered: https://github.com/ros2/rmw_connext/pull/258/files#diff-3bcd912088bec36df10ef09cd4c564b1R220

gaoethan:
@Karsten1987 @dirk-thomas yes, you're correct and I missed that, thanks :+1:

PR #266, opened 2017-10-29
Logging manipulation API from rmw_connext

sriramster:
This follows a series of changes on all the DDS specific implementation to let user programs manipulate logging levels. 

Connect to ros2/rmw#124


rmw_connext_cpp/src/rmw_logging.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include "rmw/rmw.h"
+
+extern "C"
+{
+// TODO(sriramster): Check support for log level manipulation with connext
+rmw_ret_t
+rmw_set_log_severity(rmw_log_severity_t severity)
+{
+  return RMW_RET_OK;
```
dirk-thomas: As long as this is not implemented it shouldn't return "ok".

sriramster: Since, this is not implemented yet. An error returned might disturb the normal execution, so for time being till we find ways to implement the same inside DDS thought a "OK" (by-pass) would be better.

dirk-thomas: Can you look into implementing this function or do propose to merge this as-is?

sriramster:
@dirk-thomas @dhood Could you please review this patch? This is connected to another approved patch by Dirk

dirk-thomas:
> Could you please review this patch?

I am still waiting for an answer on my [previous question](https://github.com/ros2/rmw_connext/pull/266#discussion_r149206104).

rmw_connext_cpp/src/rmw_logging.cpp
sriramster: @dirk-thomas Please merge as-is. I'll look into implementing this patch at later point. Thank you.

PR #274, opened 2017-12-08
Only set NDDSHOME if not previously set

dhood:


dhood:
this will need another iteration based on discussion in https://github.com/ros2/rmw_opensplice/pull/150#issuecomment-380177918. I will update this one after https://github.com/ros2/rmw_opensplice/pull/219 has been reviewed.

dhood:
in review; have tested on all platforms and works as expected.

On OS X I had connext for an older platform which is what prompted [this check](https://github.com/ros2/rmw_connext/pull/274/files#diff-26d02455d756338f7181f63e4981b596R40)

BTW, FWIU we don't just source the setup script in the sh file (like we do in the batch file) because the bash file connext provides doesn't work with sh out-of-the-box.

dhood:
CI including https://github.com/ros2/rmw_opensplice/pull/229 (but manual testing was done with packaging jobs that included connext experimentally)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/4690/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/1577/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/3889/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/4768/)

PR #276, opened 2018-01-19
Node name in user data

dirk-thomas:
Connect to ros2/ros2#438.

### DIFF
--- a/rmw_connext_shared_cpp/src/node.cpp
+++ b/rmw_connext_shared_cpp/src/node.cpp
@@ -53,6 +53,23 @@ create_node(
   // This String_dup is not matched with a String_free because DDS appears to
   // free this automatically.
   participant_qos.participant_name.name = DDS::String_dup(name);
+  // since the participant name is not part of the DDS spec
+  // the node name is also set in the user_data
+  DDS_Long name_length = static_cast<DDS_Long>(strlen(name));
+  const char prefix[6] = "name=";
+  bool success = participant_qos.user_data.value.length(name_length + sizeof(prefix));
+  if (!success) {
+    RMW_SET_ERROR_MSG("failed to resize participant user_data");
+    return NULL;
+  }
+  memcpy(participant_qos.user_data.value.get_contiguous_buffer(), prefix, sizeof(prefix) - 1);
+  {
+    for (DDS_Long i = 0; i < name_length; ++i) {
+      participant_qos.user_data.value[sizeof(prefix) - 1 + i] = name[i];
+    }
+    participant_qos.user_data.value[sizeof(prefix) - 1 + name_length] = ';';
+  }
+
   // forces local traffic to be sent over loopback,
   // even if a more efficient transport (such as shared memory) is installed
   // (in which case traffic will be sent over both transports)
--- a/rmw_connext_shared_cpp/src/node_names.cpp
+++ b/rmw_connext_shared_cpp/src/node_names.cpp
@@ -12,11 +12,15 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <string>
+#include <vector>
+
 #include "rcutils/logging_macros.h"
 #include "rcutils/strdup.h"
 
 #include "rmw/convert_rcutils_ret_to_rmw_ret.h"
 #include "rmw/error_handling.h"
+#include "rmw/impl/cpp/key_value.hpp"
 #include "rmw/sanity_checks.h"
 
 #include "rmw_connext_shared_cpp/ndds_include.hpp"
@@ -67,15 +71,31 @@ get_node_names(
     return RMW_RET_BAD_ALLOC;
   }
 
-
   for (auto i = 1; i < length; ++i) {
     DDS::ParticipantBuiltinTopicData pbtd;
     auto dds_ret = participant->get_discovered_participant_data(pbtd, handles[i - 1]);
-    const char * name = pbtd.participant_name.name;
-    if (!name || dds_ret != DDS_RETCODE_OK) {
-      name = "(no name)";
+    std::string name;
+    if (DDS_RETCODE_OK == dds_ret) {
+      auto data = static_cast<unsigned char *>(pbtd.user_data.value.get_contiguous_buffer());
+      std::vector<uint8_t> kv(data, data + pbtd.user_data.value.length());
+      auto map = rmw::impl::cpp::parse_key_value(kv);
+      auto found = map.find("name");
+      if (found != map.end()) {
+        name = std::string(found->second.begin(), found->second.end());
+      }
+      if (name.empty()) {
+        // use participant name if no name was found in the user data
+        if (pbtd.participant_name.name) {
+          name = pbtd.participant_name.name;
+        }
+      }
     }
-    node_names->data[i] = rcutils_strdup(name, allocator);
+    if (name.empty()) {
+      // ignore discovered participants without a name
+      node_names->data[i] = nullptr;
+      continue;
+    }
+    node_names->data[i] = rcutils_strdup(name.c_str(), allocator);
     if (!node_names->data[i]) {
       RMW_SET_ERROR_MSG("could not allocate memory for node name")
       rcutils_ret = rcutils_string_array_fini(node_names);


rmw_connext_shared_cpp/src/node.cpp
```diff
@@ -53,6 +53,23 @@ create_node(
...
   participant_qos.participant_name.name = DDS::String_dup(name);
+  // since the participant name is not part of the DDS spec
+  // the node name is also set in the user_data
+  DDS_Long name_length = strlen(name);
+  bool success = participant_qos.user_data.value.length(name_length + 6);
+  if (!success) {
+    RMW_SET_ERROR_MSG("failed to resize participant user_data");
+    return NULL;
+  }
+  const char * prefix = "name=";
```
mikaelarguedas: We would reduce the number of hardcoded values if we defined this prefix above and use it's length everywhere below (rather than hardcoding `5`s and `6`s)

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -67,15 +71,28 @@ get_node_names(
...
 
-
   for (auto i = 1; i < length; ++i) {
     DDS::ParticipantBuiltinTopicData pbtd;
     auto dds_ret = participant->get_discovered_participant_data(pbtd, handles[i - 1]);
-    const char * name = pbtd.participant_name.name;
-    if (!name || dds_ret != DDS_RETCODE_OK) {
-      name = "(no name)";
+    std::string name;
+    if (dds_ret == DDS_RETCODE_OK) {
```
mikaelarguedas: Nit: enums and litterals on left side of operators

rmw_connext_shared_cpp/src/node.cpp
dirk-thomas: Done in 95dfd983a9f62769c98a6e9af5b47c0beddf27e7.

rmw_connext_shared_cpp/src/node_names.cpp
dirk-thomas: Done in 93e4877c3d699d5360ebd9a1889a1d2c3564af1e.

rmw_connext_shared_cpp/src/node.cpp
dirk-thomas: Done in 221769e5d58d3c778a239a71039226b183063efe.

rmw_connext_shared_cpp/src/node_names.cpp
dirk-thomas: Done in 1164ceae56fc7f507ea91c64ee2a4acd8678d592.

rmw_connext_shared_cpp/src/node.cpp
```diff
@@ -53,6 +53,23 @@ create_node(
   // This String_dup is not matched with a String_free because DDS appears to
   // free this automatically.
   participant_qos.participant_name.name = DDS::String_dup(name);
+  // since the participant name is not part of the DDS spec
+  // the node name is also set in the user_data
+  DDS_Long name_length = strlen(name);
```
mikaelarguedas: This generates a [warning on Windows](http://ci.ros2.org/job/ci_windows/4009/warnings41Result/new/)

dirk-thomas: Fixed.

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -67,15 +71,29 @@ get_node_names(
...
+      auto data = static_cast<unsigned char *>(pbtd.user_data.value.get_contiguous_buffer());
+      std::vector<uint8_t> kv(data, data + pbtd.user_data.value.length());
+      auto map = rmw::impl::cpp::parse_key_value(kv);
+      auto found = map.find("name");
+      if (found != map.end()) {
+        name = std::string(found->second.begin(), found->second.end());
+      }
+      if (name.empty()) {
+        // use participant name if no name was found in the user data
+        name = pbtd.participant_name.name;
```
dhood: The participant name might be null (that seems to be the case for whatever additional participants opensplice is spawning). If we just store the participant name without checking, this causes a segfault at the `if (name.empty())` check on the `std::string` [below](https://github.com/ros2/rmw_connext/pull/276/files#diff-bccfea1954134a50cf75b01ae95188e3R91)

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -67,15 +71,29 @@ get_node_names(
...
+      }
+    }
+    if (name.empty()) {
+      // ignore discovered participants without a name
+      node_names->data[i] = nullptr;
+      continue;
```
dhood: I don't think `continue`ing here is a good idea, higher layers are probably assuming that the `node_names` map will return valid strings (nothing in the `rcl_get_node_names` documentation suggests otherwise). FWICT `rclpy`, for example, is assuming null-terminated strings are stored at each index: https://github.com/ros2/rclpy/blob/fbd80a27342d56b32047ecb7f2ad2e44e9d027b3/rclpy/src/rclpy/_rclpy.c#L2323

rmw_connext_shared_cpp/src/node_names.cpp
dirk-thomas: That makes sense. While the participant name is a `std::string` in other implementations it is a `char *` in Connext. I checked for `NULL` in ae30090bcfacb88081f4031c8965493e654347fd before assigning it to a `std::string`.

rmw_connext_shared_cpp/src/node_names.cpp
dirk-thomas: The line you refenced should be fine if `NULL` is passed in as far as I understand the API docs:

> If u is NULL, this function behaves like PyUnicode_FromUnicode() with the buffer set to NULL. This usage is deprecated in favor of PyUnicode_New().

dirk-thomas:
> Are you testing this with opensplice 6.7 or a different version?

Yes, with 6.7.

> When I tested the behaviour with ros2 node list cross-vendor (to verify ros2/ros2cli#72 is resolved), it didn't work. A connext daemon will detect an opensplice talker node now, but still has issues with whatever other participants opensplice spawns.

Can you describe what exact issues you are seeing. Maybe with a sequence of commands you are running in order to reproduce it.

rmw_connext_shared_cpp/src/node_names.cpp
dhood: Looks like you're referencing the documentation for `PyUnicode_FromStringAndSize`, `PyUnicode_FromString` [says](https://docs.python.org/3/c-api/unicode.html):

> Create a Unicode object from a UTF-8 encoded null-terminated char buffer u.

The error (when it doesn't segfault) is: `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xf0 in position 0: invalid continuation byte`

dhood:
The specific issue is the segfault that I mentioned in the inline comments.

The sequence is the same as in the initial issue:
```
RMW_IMPLEMENTATION=rmw_opensplice_cpp ros2 run demo_nodes_cpp talker
```
```
$ RMW_IMPLEMENTATION=rmw_connext_cpp ros2 node list
RTI Data Distribution Service EVAL License issued to OSRF dhood@osrfoundation.org For non-production use only.
Expires on 05-Mar-2018 See www.rti.com for more information.
Segmentation fault (core dumped)

```

dhood:
What output are you seeing/how are you testing the cross-vendor behaviour? Does `ros2 node list` have blank lines still? Maybe it's using an older version of the daemon instead of a "fresh" one?

rmw_connext_shared_cpp/src/node_names.cpp
dirk-thomas: You are right, I stopped at the first match and didn't check the exact function name :disappointed:

dirk-thomas:
I was able to reproduce the problem with the OpenSplice talker and the Connext node list. I created PRs to document the behavior of the [rcl](https://github.com/ros2/rcl/pull/214) API as well as fix the usage in [rclcpp](https://github.com/ros2/rclcpp/pull/435) and [rclpy](https://github.com/ros2/rclpy/pull/177).

Two considered alternatives:
* Each rmw implementation could implement `rmw_get_node_names()` with a two-pass: the first iteration only checks how many non-null names exist, then it allocates the array with the exact size, in the second iteration it populates the array with all non-null names.
* If the `rcutils_string_array_t` struct would have a `capacity` member storing the number of array items which actually contain data (beside the existing allocated `size`) `null` node names could be skipped in each rmw implementation already in a single pass without the need for reallocating the array.

PR #277, opened 2018-02-12
compile all cpp files

Karsten1987:
I've noticed that a couple of `.cpp` files in the `src` folder are currently not compiled. As our CI turns out to run just fine, I think these functions are currently not being used anywhere.
Don't know exactly why, but I think I simply forgot them to add when restructuring this package.

I don't run CI/linters on this PR since the files are still taken into considerations by the linters.

### DIFF
--- a/connext_cmake_module/cmake/Modules/FindConnext.cmake
+++ b/connext_cmake_module/cmake/Modules/FindConnext.cmake
@@ -313,6 +313,8 @@ if(Connext_FOUND)
   endif()
 
   if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+    set(Connext_LIBRARIES "-Wl,--no-as-needed" ${Connext_LIBRARIES} "-Wl,--as-needed")
+
     # check with which ABI the Connext libraries are built
     configure_file(
       "${connext_cmake_module_DIR}/check_abi.cmake"
--- a/connext_cmake_module/cmake/check_abi.cmake
+++ b/connext_cmake_module/cmake/check_abi.cmake
@@ -15,7 +15,7 @@
 cmake_minimum_required(VERSION 2.8.3)
 
 if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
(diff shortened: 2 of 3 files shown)


mikaelarguedas:
Should we run CI (without linters) to make sure that these files compile without warnings on all platforms even if they're not used by downstream packages?

Karsten1987:
good point!
CI:

linux: [Build Status](http://ci.ros2.org/job/ci_linux/3957/)
osx: [Build Status](http://ci.ros2.org/job/ci_osx/3292/)
win: [Build Status](http://ci.ros2.org/job/ci_windows/4059/)

Karsten1987:
Linux is currently failing with
`
/home/rosbuild/rti_connext_dds-5.3.0/lib/x64Linux3gcc5.4.0/libnddscore.so: undefined reference to 'dlopen'
`
I'll investigate why that occurs only on the linux job.

dhood:
you can try reverting https://github.com/ros2/rmw_connext/pull/275

Karsten1987:
nice catch! Reverting #275 seems to work. Any opinions on this? According to the compilation error, it seems we need to link against `dl`. Does anybody know which libraries are set on top of this in 275? I haven't really looked into it.

linux: [Build Status](http://ci.ros2.org/job/ci_linux/3959/)
osx: [Build Status](http://ci.ros2.org/job/ci_osx/3293/)
win: [Build Status](http://ci.ros2.org/job/ci_windows/4060/)

mikaelarguedas:
I thinks that'd be fine to revert 275. Based on the connext release notes that didnt seem to be necessary anymore but apparently it still is. May be worth checking the latest release notes to make sure and ping them to know if that'll be fixed in the next patch version.

For the record ldd on `libnddscore.so` returns: `/lib64/ld-linux-x86-64.so.2 (0x00007fc200986000)`

nuclearsandwich:
linux: [Build Status](https://ci.ros2.org/job/ci_linux/3959/)
osx: [Build Status](https://ci.ros2.org/job/ci_osx/3293/)
win: [Build Status](https://ci.ros2.org/job/ci_windows/4060/)

Testing build status embedding with https

PR #285, opened 2018-04-13
Remove topic partitions

rohitsalem:
connects to ros2/ros2#476 
To use the topic/service name directly without using the partitions.

### DIFF
--- a/rmw_connext_cpp/src/process_topic_and_service_names.cpp
+++ b/rmw_connext_cpp/src/process_topic_and_service_names.cpp
@@ -27,51 +27,26 @@ bool
 _process_topic_name(
   const char * topic_name,
   bool avoid_ros_namespace_conventions,
-  char ** topic_str,
-  char ** partition_str)
+  char ** topic_str)
 {
   bool success = true;
-  rcutils_string_array_t name_tokens = rcutils_get_zero_initialized_string_array();
   rcutils_allocator_t allocator = rcutils_get_default_allocator();
 
-  if (rcutils_split_last(topic_name, '/', allocator, &name_tokens) != RCUTILS_RET_OK) {
-    RMW_SET_ERROR_MSG(rcutils_get_error_string_safe())
-    success = false;
-    goto end;
-  }
-  if (name_tokens.size == 1) {
-    if (!avoid_ros_namespace_conventions) {
-      *partition_str = DDS_String_dup(ros_topic_prefix);
+  if (!avoid_ros_namespace_conventions) {
+    char * concat_str =
+      rcutils_format_string(allocator, "%s%s", ros_topic_prefix, topic_name);
+    if (!concat_str) {
+      RMW_SET_ERROR_MSG("could not allocate memory for topic string")
+      success = false;
+      goto end;
     }
-    *topic_str = DDS_String_dup(name_tokens.data[0]);
-  } else if (name_tokens.size == 2) {
-    if (avoid_ros_namespace_conventions) {
-      // no ros_topic_prefix, so store the user's namespace directly
-      *partition_str = DDS_String_dup(name_tokens.data[0]);
-    } else {
-      // concat the ros_topic_prefix with the user's namespace
-      char * concat_str =
-        rcutils_format_string(allocator, "%s/%s", ros_topic_prefix, name_tokens.data[0]);
-      if (!concat_str) {
-        RMW_SET_ERROR_MSG("could not allocate memory for partition string")
-        success = false;
-        goto end;
-      }
-      *partition_str = DDS_String_dup(concat_str);
-      allocator.deallocate(concat_str, allocator.state);
-    }
-    *topic_str = DDS_String_dup(name_tokens.data[1]);
+    *topic_str = DDS_String_dup(concat_str);
+    allocator.deallocate(concat_str, allocator.state);
   } else {
-    RMW_SET_ERROR_MSG("incorrectly formatted topic name")
-    success = false;
+    *topic_str = DDS_String_dup(topic_name);
   }
 
 end:
-  // all necessary strings are copied into connext
-  // free that memory
-  if (rcutils_string_array_fini(&name_tokens) != RCUTILS_RET_OK) {
-    fprintf(stderr, "Failed to destroy the token string array\n");
-  }
   return success;
 }
 
--- a/rmw_connext_cpp/src/process_topic_and_service_names.cpp
+++ b/rmw_connext_cpp/src/process_topic_and_service_names.cpp
@@ -54,64 +54,63 @@ bool
 _process_service_name(
   const char * service_name,
   bool avoid_ros_namespace_conventions,
-  char ** service_str,
-  char ** request_partition_str,
-  char ** response_partition_str)
+  char ** request_topic_str,
+  char ** response_topic_str)
 {
   bool success = true;
-  rcutils_string_array_t name_tokens = rcutils_get_zero_initialized_string_array();
+
   rcutils_allocator_t allocator = rcutils_get_default_allocator();
 
-  if (rcutils_split_last(service_name, '/', allocator, &name_tokens) != RCUTILS_RET_OK) {
-    RMW_SET_ERROR_MSG(rcutils_get_error_string_safe())
-    success = false;
-    goto end;
-  }
-  if (name_tokens.size == 1) {
-    if (!avoid_ros_namespace_conventions) {
-      *request_partition_str = DDS_String_dup(ros_service_requester_prefix);
-      *response_partition_str = DDS_String_dup(ros_service_response_prefix);
-    }
-    *service_str = DDS_String_dup(name_tokens.data[0]);
-  } else if (name_tokens.size == 2) {
-    if (avoid_ros_namespace_conventions) {
-      // no ros_service_*_prefix, so store the user's namespace directly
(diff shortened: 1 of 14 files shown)


rmw_connext_shared_cpp/src/node.cpp
```diff
@@ -70,6 +70,9 @@ create_node(
     participant_qos.user_data.value[sizeof(prefix) - 1 + name_length] = ';';
   }
 
+  // Setting the contentfilter_property_max_length to 1024
```
mikaelarguedas: Can you please modify the comment to describe why it's needed and how 1024 has been picked?

rmw_connext_shared_cpp/src/node.cpp
```diff
@@ -70,6 +70,9 @@ create_node(
...
+  participant_qos.resource_limits.contentfilter_property_max_length = 1024;
```
mikaelarguedas: As it been confirm that we can still talk to native connext participants if we modify this default value ?

rmw_connext_shared_cpp/src/node.cpp
rohitsalem: Yes, I'm still investigating the exact number that should be used for the property, will update once I have a reasonable one.

rmw_connext_shared_cpp/src/node.cpp
rohitsalem: Still investigating that to confirm along with the exact length for the property.

rohitsalem: I've checked native connext replier and requester, changed the `contentfilter_property_max_length` of requester to be 1024, replier to be 256 and vice-versa. The replier and requester worked as expected.

rmw_connext_shared_cpp/src/node.cpp
mikaelarguedas: for reference: https://community.rti.com/static/documentation/connext-dds/5.3.1/doc/api/connext_dds/api_cpp/structDDS__DomainParticipantResourceLimitsQosPolicy.html#aef1ff851202f5777d31528a52d528e4f

From the RTPS spec:
```
ContentFilterProperty_t: Mapping of the type

typedef string<256> String256;
typedef sequence<string> StringSequence;
struct ContentFilterProperty_t {
  String256 contentFilteredTopicName;
  String256 relatedTopicName;
  String256 filterClassName;
  string filterExpression;
  StringSequence expressionParameters;
};
```

mikaelarguedas: yes I think it makes sense to have the link in the comment as well

mikaelarguedas:
As we found out that the service length has to be shorter with Connext. Can you post here what the error message is if we pass a service name within "ROS maximum length" but too long for Connext?
It may be worth add a check here to return a sensible error message in the case where it's too long

rohitsalem:
Here is the error message when I run `ros2 run demo_nodes_cpp add_two_ints_client` when the service_name is within "ROS maximum length"

```
rsalem@warner:~/work/ros2_ws$ ros2 run demo_nodes_cpp add_two_ints_client
RTI Data Distribution Service Evaluation License issued to OSRF dthomas@osrfoundation.org For non-production use only.
Expires on 05-may-2018 See www.rti.com for more information.
CorrelationCFTBuilder::create_correlation_cft:ERROR: Bad parameter: topic name is too long: rr/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaReply
connext::details::EntityUntypedImpl::initialize:!failed (see previous errors)

>>> [rcutils|error_handling.c:155] rcutils_set_error_state()
This error state is being overwritten:

  'C++ exception during construction of Requester, at /home/rsalem/work/ros2_ws/build_isolated/example_interfaces/rosidl_typesupport_connext_cpp/example_interfaces/srv/dds_connext/add_two_ints__type_support.cpp:100'

with this new error message:

  'failed to create requester, at /home/rsalem/work/ros2_ws/src/ros2/rmw_connext/rmw_connext_cpp/src/rmw_client.cpp:141'

rcutils_reset_error() should be called after error handling to avoid this.
<<<
terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
  what():  could not create client: failed to create requester, at /home/rsalem/work/ros2_ws/src/ros2/rmw_connext/rmw_connext_cpp/src/rmw_client.cpp:141, at /home/rsalem/work/ros2_ws/src/ros2/rcl/rcl/src/rcl/client.c:174
```

rohitsalem:
regarding the check for the length of the service_name, I think It should be added in [_process_service_name](https://github.com/ros2/rmw_connext/blob/master/rmw_connext_cpp/src/process_topic_and_service_names.cpp#L79) function, but still as of now what should the max limit for service_name in connext should be? When I checked the native connext Requester Replier example, the maximum service name that I was able to use was of 185 characters, If we have to impose a limit here I think it should be 185U - 8U (to account for the ros prefixes)

mikaelarguedas:
Actually as the service length is not checked properly at the moment for any rmw implementation (we check the length before adding Request or Reply to the topic names and don't account for it), let's keep that for another set of PRs to make sure these ones get reviewed / merged in a timely manner.

Can you open an issue on rmw to relocate the discussion there ?

rmw_connext_cpp/src/process_topic_and_service_names.cpp
```diff
@@ -27,116 +27,90 @@ bool
...
-        success = false;
-        goto end;
-      }
-      *request_partition_str = DDS_String_dup(request_concat_str);
-      *response_partition_str = DDS_String_dup(response_concat_str);
+  if (!avoid_ros_namespace_conventions) {
+    // concat the ros_service_*_prefix and Request/Reply suffixes with the service_name
+    char * request_concat_str = rcutils_format_string(
+      allocator,
+      "%s%s%s", ros_service_requester_prefix, service_name, "Request");
```
mikaelarguedas: hmmm Do we need to add Request and Reply ourselves now ? Isn't Connext adding them automatically when we call their service API ?

rmw_connext_cpp/src/process_topic_and_service_names.cpp
```diff
@@ -27,116 +27,90 @@ bool
...
+    if (!request_concat_str) {
       allocator.deallocate(request_concat_str, allocator.state);
+      RMW_SET_ERROR_MSG("could not allocate memory for request topic string")
+      success = false;
+      goto end;
+    }
+    char * response_concat_str = rcutils_format_string(
+      allocator,
+      "%s%s%s", ros_service_response_prefix, service_name, "Reply");
+    if (!response_concat_str) {
```
Karsten1987: i think `request_concat_str` is leaked here. If it was previously allocated, but `response_concat_str` fails here, only the request string will be deallocated.

rmw_connext_cpp/src/process_topic_and_service_names.cpp
```diff
@@ -27,116 +27,90 @@ bool
...
-        "%s/%s", ros_service_response_prefix, name_tokens.data[0]);
-      if (!response_concat_str) {
-        allocator.deallocate(request_concat_str, allocator.state);
-        RMW_SET_ERROR_MSG("could not allocate memory for partition string")
-        success = false;
-        goto end;
-      }
-      *request_partition_str = DDS_String_dup(request_concat_str);
-      *response_partition_str = DDS_String_dup(response_concat_str);
+  if (!avoid_ros_namespace_conventions) {
```
Karsten1987: Consider to simply set the ros prefix to an empty string and run the same logic when concatenating the topic string. Because if I understand your code correctly, the only difference is whether you append the ros prefix or not, right?
This makes the logic for cleaning up the allocated strings simpler.


``` pseudo
string _ros_prefix = ros_prefix;
if (avoid_ros_namespace_conventions) {
   _ros_prefix = "";
}

topic = (%s%s%s, _ros_prefix, topic_name, "[Request|Reply]"
```

rmw_connext_cpp/src/rmw_client.cpp
```diff
@@ -129,12 +127,16 @@ rmw_create_client(
   // This has to be evaluated whether or not to provide a
   // Subscriber/Publisher object directly with preset partitions.
```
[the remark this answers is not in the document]
rohitsalem: @Karsten1987  [this](https://github.com/ros2/rmw_connext/pull/285/files/c23364791279b960eacbf1dbc03609e0fd41fe36#diff-58ed3a42616228c76e99789cb4c3f88cL127) is a TODO, can it be removed or should it be modified?

rmw_connext_cpp/src/process_topic_and_service_names.cpp
rohitsalem: @Karsten1987 I think I got the logic wrong. So should it be if the `response_concat_str` fails here, the `request_concat_str` for which memory was set previously is deallocated? I see that's what was there previously

PR #286, opened 2018-04-16
add COLCON_IGNORE file

dirk-thomas:
This patch serves as a template for review. Once it got approved I would commit similar changes to the following repos / packages (without the overhead of a PR for each):

* `geometry2/geometry_experimental`
* `geometry2/geometry2`
* `geometry2/test_tf2`
* `geometry2/tf2_bullet`
* `geometry2/tf2_kdl`
* `geometry2/tf2_py`
* `geometry2/tf2_sensor_msgs`
* `geometry2/tf2_tools`
* `kdl_parser/kdl_parser_py`
* `orocos_kinematics_dynamics/python_orocos_kdl`
* `rviz/rviz`
* `tutorials/rclc_tutorials`
* `urdf/urdf_parser_plugin`

I will post links to the commits once that is done.


mikaelarguedas:
Is the goal to keep both the `COLCON_IGNORE` and `AMENT_IGNORE` files in the future to support the use of both tools (colcon and ament_tools) ?

dirk-thomas:
For now I would just add the `COLCON_IGNORE` files. And once the transition is complete, we have updated all the instructions etc. we can remove the `AMENT_IGNORE` files. The goal is to finish the transition for Bouncy.

dirk-thomas:
Since the `colcon-ros` package also ignores trees with an `AMENT_IGNORE` files (see https://github.com/colcon/colcon-ros/blob/2659c28746a9f3f0499b331555b767c8517361d2/colcon_ros/package_identification/ros.py#L52) this shouldn't be necessary.

PR #294, opened 2018-06-21
Use key rti-connext-dds as package key name for connext.

nuclearsandwich:
Since all connext support is provided by the one package, use a single rosdep key which matches the name of the provided package.

Do we want to add some portion of the package version to the key name as we do in https://github.com/ros2/rmw_opensplice/pull/232 in order to allow for different connext versions both on bionic?

### DIFF
--- a/rmw_connext_cpp/package.xml
+++ b/rmw_connext_cpp/package.xml
@@ -14,22 +14,22 @@
   <buildtool_export_depend>rosidl_cmake</buildtool_export_depend>
 
   <build_depend>connext_cmake_module</build_depend>
-  <build_depend>libndds52-dev</build_depend>
   <build_depend>rcutils</build_depend>
   <build_depend>rmw</build_depend>
   <build_depend>rmw_connext_shared_cpp</build_depend>
   <build_depend>rosidl_generator_c</build_depend>
   <build_depend>rosidl_generator_cpp</build_depend>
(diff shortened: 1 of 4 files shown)


rmw_connext_cpp/package.xml
```diff
@@ -14,22 +14,22 @@
   <buildtool_export_depend>rosidl_cmake</buildtool_export_depend>
 
   <build_depend>connext_cmake_module</build_depend>
-  <build_depend>libndds52-dev</build_depend>
   <build_depend>rcutils</build_depend>
   <build_depend>rmw</build_depend>
   <build_depend>rmw_connext_shared_cpp</build_depend>
   <build_depend>rosidl_generator_c</build_depend>
   <build_depend>rosidl_generator_cpp</build_depend>
+  <build_depend>rosidl_generator_dds_idl</build_depend>
```
nuclearsandwich: This move (from L25) comes from alpha-sorting the dependency list. If desired we can split it out into its own commit or remove it from the PR entirely.

dirk-thomas:
Since the actual Debian package name is `rti-connext-dds-5.3.1` I would use the exact same name as the rosdep key.

mikaelarguedas:
:+1: for using the full name as we want to be able to track the version in case we ship a different version in bionic in the future

PR #299, opened 2018-08-28
Include node namespaces in get_node_names

mjcarroll:
Connects to ros2/rmw#142

### DIFF
--- a/rmw_connext_shared_cpp/src/node.cpp
+++ b/rmw_connext_shared_cpp/src/node.cpp
@@ -63,10 +63,10 @@ create_node(
     return NULL;
   }
 
-  size_t written =
+  int written =
     snprintf(reinterpret_cast<char *>(participant_qos.user_data.value.get_contiguous_buffer()),
       length, "name=%s;namespace=%s;", name, namespace_);
-  if (written < 0) {
+  if (written < 0 || written > static_cast<int>(length) - 1) {
     RMW_SET_ERROR_MSG("failed to populate user_data buffer");
     return NULL;
   }
--- a/rmw_connext_shared_cpp/src/node.cpp
+++ b/rmw_connext_shared_cpp/src/node.cpp
@@ -55,21 +55,16 @@ create_node(
   participant_qos.participant_name.name = DDS::String_dup(name);
   // since the participant name is not part of the DDS spec
   // the node name is also set in the user_data
-  DDS_Long name_length = static_cast<DDS_Long>(strlen(name));
-  const char prefix[6] = "name=";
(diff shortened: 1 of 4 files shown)


rmw_connext_shared_cpp/src/node.cpp
```diff
@@ -55,21 +55,16 @@ create_node(
...
   }
-  memcpy(participant_qos.user_data.value.get_contiguous_buffer(), prefix, sizeof(prefix) - 1);
-  {
-    for (DDS_Long i = 0; i < name_length; ++i) {
-      participant_qos.user_data.value[sizeof(prefix) - 1 + i] = name[i];
-    }
-    participant_qos.user_data.value[sizeof(prefix) - 1 + name_length] = ';';
-  }
 
+  snprintf(reinterpret_cast<char *>(participant_qos.user_data.value.get_contiguous_buffer()),
```
wjwwood: Might be a good idea to check the return code of `snprintf`.

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -93,12 +110,28 @@ get_node_names(
...
+
     node_names->data[i] = rcutils_strdup(name.c_str(), allocator);
     if (!node_names->data[i]) {
       RMW_SET_ERROR_MSG("could not allocate memory for node name")
       rcutils_ret = rcutils_string_array_fini(node_names);
+      rcutils_ret = rcutils_string_array_fini(node_namespaces);
+      if (rcutils_ret != RCUTILS_RET_OK) {
+        RCUTILS_LOG_ERROR_NAMED(
+          "rmw_connext_cpp",
+          "failed to cleanup during error handling: %s", rcutils_get_error_string_safe())
```
wjwwood: Need to reset the error message here with `rcutils_reset_error()`.

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -93,12 +110,28 @@ get_node_names(
...
+      }
+      return RMW_RET_BAD_ALLOC;
+    }
+
+    node_namespaces->data[i] = rcutils_strdup(namespace_.c_str(), allocator);
+    if (!node_namespaces->data[i]) {
+      RMW_SET_ERROR_MSG("could not allocate memory for node namespace")
+      rcutils_ret = rcutils_string_array_fini(node_names);
+      rcutils_ret = rcutils_string_array_fini(node_namespaces);
```
wjwwood: If both fail then you'll need to reset the error state between them, but then again, perhaps that's ok or even intentional.

Separately I'll say that rather than duplicating this cleanup code, I'd recommend using a label and a goto to have it happen in one place.

PR #303, opened 2018-09-23
Straightforward approach for not returning non-named nodes

CaptainTrunky:
Connects to ros2/ros2#489

A pull request for [#489](https://github.com/ros2/ros2/issues/489) basic implementation that allocates a temporary buffer for node names and filters out all empty names. It's not properly tested yet and additional unit tests are required. So next steps are:

- add tests

- add optimal implementation that does not require additional buffer and copying data


rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -65,12 +65,19 @@ get_node_names(
...
+  rcutils_ret = rcutils_string_array_init(&tmp_names_list, length, &allocator);
+
+  tmp_names_list.data[0] = rcutils_strdup(participant_qos.participant_name.name, allocator);
+
+  if (!tmp_names_list.data[0]) {
     RMW_SET_ERROR_MSG("could not allocate memory for node name")
     return RMW_RET_BAD_ALLOC;
   }
 
+  int named_nodes_num = 0;
```
dhood: doesn't this need to start at 1 since we've already added the 0th entry?

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -65,12 +65,19 @@ get_node_names(
...
-  node_names->data[0] = rcutils_strdup(participant_qos.participant_name.name, allocator);
-  if (!node_names->data[0]) {
+
+  rcutils_string_array_t tmp_names_list = rcutils_get_zero_initialized_string_array();
+  rcutils_ret = rcutils_string_array_init(&tmp_names_list, length, &allocator);
+
+  tmp_names_list.data[0] = rcutils_strdup(participant_qos.participant_name.name, allocator);
+
+  if (!tmp_names_list.data[0]) {
     RMW_SET_ERROR_MSG("could not allocate memory for node name")
```
dhood: `rcutils_string_array_fini` would need to be called here

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -65,12 +65,19 @@ get_node_names(
     RMW_SET_ERROR_MSG("failed to get default participant qos");
     return RMW_RET_ERROR;
   }
-  node_names->data[0] = rcutils_strdup(participant_qos.participant_name.name, allocator);
-  if (!node_names->data[0]) {
+
+  rcutils_string_array_t tmp_names_list = rcutils_get_zero_initialized_string_array();
+  rcutils_ret = rcutils_string_array_init(&tmp_names_list, length, &allocator);
```
dhood: I know this might be missing in other parts of our code base, but we should check here that this returns `RCUTILS_RET_OK` (as is done [here](https://github.com/ros2/rmw_connext/pull/303/files#diff-bccfea1954134a50cf75b01ae95188e3R57))

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -90,23 +97,40 @@ get_node_names(
...
           "rmw_connext_cpp",
           "failed to cleanup during error handling: %s", rcutils_get_error_string_safe())
       }
       return RMW_RET_BAD_ALLOC;
     }
+
+    ++named_nodes_num;
+  }
+
+  rcutils_ret = rcutils_string_array_init(node_names, named_nodes_num, &allocator);
```
dhood: since we initialise this here with the required size, we can remove [this previous initialisation](https://github.com/ros2/rmw_connext/pull/303/files#diff-bccfea1954134a50cf75b01ae95188e3R56). Removing it above also means we won't have to fini the `node_names` array in the various error cases.

dhood:
FYI (since I think this PR is based off https://github.com/ros2/rmw_opensplice/pull/191): this is the PR adapting to the new API of `get_node_names` in `rmw_opensplice_cpp`: https://github.com/ros2/rmw_opensplice/pull/243/files

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: fixed

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: fixed. yes, I think that's better

CaptainTrunky:
I think everything mentioned is fixed now. Next step is adding some tests.

nuclearsandwich:
@CaptainTrunky would you be able to add tests and update this PR?

CaptainTrunky:
> @CaptainTrunky would you be able to add tests and update this PR?

Sorry for being late. I will try to finish within next few weeks.

CaptainTrunky:
I've merged actual code base and opened a corresponding PR ros2/system_tests#374

PR #308, opened 2018-11-19
refactor to support init options and context

wjwwood:
Connects to ros2/rmw#154

@serge-nikulin FYI this is an example of what the disruption to an existing rmw implementation will look like for the init options and context changes.

### DIFF
--- a/rmw_connext_cpp/src/rmw_guard_condition.cpp
+++ b/rmw_connext_cpp/src/rmw_guard_condition.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "rmw/impl/cpp/macros.hpp"
 #include "rmw/rmw.h"
 
 #include "rmw_connext_shared_cpp/guard_condition.hpp"
@@ -21,8 +22,15 @@
 extern "C"
 {
 rmw_guard_condition_t *
-rmw_create_guard_condition()
+rmw_create_guard_condition(rmw_context_t * context)
 {
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context, NULL);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    init context,
+    context->implementation_identifier,
+    rti_connext_identifier,
+    // TODO(wjwwood): replace this with RMW_RET_INCORRECT_RMW_IMPLEMENTATION when refactored
+    return NULL);
   return create_guard_condition(rti_connext_identifier);
 }
 
--- a/rmw_connext_cpp/src/rmw_init.cpp
+++ b/rmw_connext_cpp/src/rmw_init.cpp
@@ -84,4 +84,18 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
   context->impl = nullptr;
   return init();
 }
+
+rmw_ret_t
+rmw_shutdown(rmw_context_t * context)
+{
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    context,
+    context->implementation_identifier,
+    rti_connext_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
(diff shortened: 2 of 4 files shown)


serge-nikulin:
Ack. 

I will also join the chorus "do we really need `rmw_init_options_t `?"

Please give me an example when a context is not enough.

wjwwood:
`rmw_init_options_t` is given as input to `init`, and `context` is an output, so if you want to influence the initialization then you need the init options.

serge-nikulin:
OK, I like it.

rmw_connext_cpp/src/rmw_guard_condition.cpp
```diff
@@ -21,8 +22,15 @@
 extern "C"
 {
 rmw_guard_condition_t *
-rmw_create_guard_condition()
+rmw_create_guard_condition(rmw_context_t * context)
 {
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context, NULL);
```
jacobperron: Could we reduce duplicate code by pushing this into `create_guard_condition`? This would require passing the context and expected identifier as arguments.

rmw_connext_cpp/src/rmw_init.cpp
```diff
@@ -12,15 +12,76 @@
...
 
+#include "rmw_connext_cpp/identifier.hpp"
+
 extern "C"
 {
 rmw_ret_t
-rmw_init()
+rmw_init_options_init(rmw_init_options_t * init_options, rcutils_allocator_t allocator)
+{
+  RMW_CHECK_ARGUMENT_FOR_NULL(init_options, RMW_RET_INVALID_ARGUMENT);
```
jacobperron: Again, could we create shared functions to be used by connext and connext_dynamic for all of these?

rmw_connext_cpp/src/rmw_guard_condition.cpp
wjwwood: I think so, but in the interest of getting this pr merged, I'd like to do that in a follow up pr. I'll make an issue on merge.

rmw_connext_cpp/src/rmw_init.cpp
```diff
@@ -12,15 +12,90 @@
...
+rmw_ret_t
+rmw_shutdown(rmw_context_t * context)
+{
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    context,
+    context->implementation_identifier,
+    rti_connext_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context->impl, RMW_RET_INVALID_ARGUMENT);
```
sloretz: It looks like `rmw_init()` sets `context->impl` to `nullptr`, why is this expecting it to be non-null?

rmw_connext_dynamic_cpp/src/functions.cpp
```diff
@@ -202,18 +203,98 @@ rmw_get_serialization_format()
...
+rmw_ret_t
+rmw_shutdown(rmw_context_t * context)
+{
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    context,
+    context->implementation_identifier,
+    rti_connext_dynamic_identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context->impl, RMW_RET_INVALID_ARGUMENT);
```
sloretz: same comment here, `context->impl` is set to `nullptr` in `rmw_init()`

serge-nikulin:
> 
> 
> Connects to [ros2/rmw#154](https://github.com/ros2/rmw/pull/154)
> 
> FYI this is an example of what the disruption to an existing rmw implementation will look like for the init options and context changes.

@wjwwood, do you have an example of a user-side customization?

serge-nikulin:
@wjwwood, 

> And so currently you cannot

Do you have an issue to address it?

wjwwood:
@serge-nikulin see: https://github.com/ros2/rcl/pull/372

PR #313, opened 2018-12-03
Node graph impl

ross-desmond:
Added node graph implementation to rti_connext

Open issue: rti_connext previously added the same topic multiple times to account for issues with discovery. When attempting to construct a node graph there seems to be an issue with matching. I'll open an issue to be addressed shortly.


ross-desmond:
Using rmw 0.6.0 in CI which does not contain rmw_context_t even though this change is tracking master. Awaiting advice as to make this change compliant with 0.6.0 and therefore create a new branch or continue with developing on master.

rmw_connext_shared_cpp/src/node_info_and_types.cpp
```diff
@@ -0,0 +1,289 @@
...
+#include "rmw/allocators.h"
+#include "rmw/convert_rcutils_ret_to_rmw_ret.h"
+#include "rmw/error_handling.h"
+#include "rmw/get_topic_names_and_types.h"
+#include "rmw/impl/cpp/macros.hpp"
+#include "rmw/impl/cpp/key_value.hpp"
+#include "rmw/names_and_types.h"
+#include "rmw/rmw.h"
+
+#include "rmw_connext_shared_cpp/types.hpp"
```
clalancette: For reasons I do not yet understand, macOS throws tons and tons of warnings on this include: https://ci.ros2.org/job/ci_osx/4759/consoleFull, search for "storage class identifier is deprecated".

wjwwood: These warnings are in Connext itself, so we suppress them in other files which include these headers, e.g.:

https://github.com/ros2/rmw_connext/blob/e8e50e12ea4a8f48201fe6947ed1cb7df8acce22/rmw_connext_shared_cpp/include/rmw_connext_shared_cpp/ndds_include.hpp#L18-L30

clalancette: Ah great, thanks.  I've fixed it in 7f7ae2d.  I still need a fix for cxx11 problem in rmw_fastrtps before I can run CI, but I'll follow up over there.

ross-desmond:
Pushed changes for windows, looks like I'm still unable to get the node_info_and_types.cpp to compile into the rmw_connext_shared_cpp.lib.

Does anyone who works with windows have any ideas?

ross-desmond:
Applying these changes for #314 does not solve the current issue. The error I am seeing is rmw_get_*_by_node functions in rmw_connext_cpp cannot find rmw_connext_shared_cpp get_*_by_node functions. These happen for the 3 new functions I've added. I've verified that the symbol in the following error are not in the rmw_connext_shared_cpp.lib in windows.


```
get_subscriber_names_and_types_by_node@@YAHPEBDPEBUrmw_node_t@@PEAUrcutils_allocator_t@@00_NPEAUrmw_names_and_types_t@@@Z) referenced in function rmw_get_subscriber_names_and_types_by_node [C:\dev\ros2\build\rmw_connext_cpp\rmw_connext_cpp.vcxproj]
C:\dev\ros2\build\rmw_connext_cpp\Release\rmw_connext_cpp.dll : fatal error LNK1120: 3 unresolved externals [C:\dev\ros2\build\rmw_connext_cpp\rmw_connext_cpp.vcxproj]
```

clalancette:
> Applying these changes for #314 does not solve the current issue. The error I am seeing is rmw_get___by_node functions in rmw_connext_cpp cannot find rmw_connext_shared_cpp get___by_node functions. These happen for the 3 new functions I've added. I've verified that the symbol in the following error are not in the rmw_connext_shared_cpp.lib in windows.

Hm, OK.  I'm debugging now on a Windows machine; I'll let you know what I find.

ross-desmond:
@clalancette Thank you for looking into it as well, we are still looking into the issue.

PR #320, opened 2018-12-13
No need to use strcmp, we can compare a std::string with operator== t…

jwillemsen:
…o a const char*, makes the code easier to read

    * rmw_connext_shared_cpp/src/node_info_and_types.cpp:


sloretz:
CI (packages up to rclcpp, only connext)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5972/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2491/) (canceled since rmw_connext doesn't support arm64)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4934/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5857/)

sloretz:
CI looks good, thanks for the PR.

PR #322, opened 2018-12-14
remove debug print

wjwwood:
Left over from: https://github.com/ros2/rmw_connext/pull/221/files#diff-becdf01440f1757df19f4775b534e066R2183


wjwwood:
This is where I discovered it:

```
$ ros2 param list
Number of response publishers is 0
Number of response publishers is 0
Number of response publishers is 0
Number of response publishers is 0
Number of response publishers is 0
Number of response publishers is 0
/talker:
```

PR #339, opened 2018-12-18
Use narrow to downcast to a type specific datareader

### LINKED ISSUE #329: Downcasting DDS entities should use dynamic_cast/narrow
#### Feature description
For downcasting DDS entities to a specific DDS entity dynamic_cast should be used, not static_cast. Even better is to use the generated narrow operations. Something for after #328 

#### Implementation considerations
Not sure whether there are code guidelines not to use dynamic_cast

jwillemsen:
Fixes #329

### DIFF
--- a/rmw_connext_shared_cpp/src/types/custom_publisher_listener.cpp
+++ b/rmw_connext_shared_cpp/src/types/custom_publisher_listener.cpp
@@ -25,6 +25,11 @@ void CustomPublisherListener::on_data_available(DDS::DataReader * reader)
   DDS::PublicationBuiltinTopicDataDataReader * builtin_reader =
     DDS::PublicationBuiltinTopicDataDataReader::narrow(reader);
 
+  if (!builtin_reader) {
+    fprintf(stderr, "failed to narrow to DDS::PublicationBuiltinTopicDataDataReader\n");
+    return;
+  }
+
   DDS::PublicationBuiltinTopicDataSeq data_seq;
   DDS::SampleInfoSeq info_seq;
   DDS::ReturnCode_t retcode = builtin_reader->take(
--- a/rmw_connext_shared_cpp/src/types/custom_publisher_listener.cpp
+++ b/rmw_connext_shared_cpp/src/types/custom_publisher_listener.cpp
@@ -23,7 +23,7 @@
 void CustomPublisherListener::on_data_available(DDS::DataReader * reader)
 {
   DDS::PublicationBuiltinTopicDataDataReader * builtin_reader =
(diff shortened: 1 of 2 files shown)


rmw_connext_shared_cpp/src/types/custom_publisher_listener.cpp
```diff
@@ -23,7 +23,7 @@
 void CustomPublisherListener::on_data_available(DDS::DataReader * reader)
 {
   DDS::PublicationBuiltinTopicDataDataReader * builtin_reader =
-    static_cast<DDS::PublicationBuiltinTopicDataDataReader *>(reader);
+    DDS::PublicationBuiltinTopicDataDataReader::narrow(reader);
```
sloretz: It looks like [narrow returns `NULL` if the cast is invalid](https://community.rti.com/kb/what-purpose-narrow-method-created-type-rtiddsgen). Mind adding a check for this case?

sloretz: Same comment on the other `narrow` call added by this PR

jwillemsen:
Addressed review comments

mjcarroll:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6020/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2526/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4971/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5895/)

rmw_connext_shared_cpp/src/types/custom_publisher_listener.cpp
```diff
@@ -23,7 +23,12 @@
...
+
+  if (!builtin_reader) {
+    fprintf(stderr, "failed to narrow to DDS::PublicationBuiltinTopicDataDataReader\n");
```
sloretz: Recommend `RCUTILS_LOG_ERROR_NAMED` instead of `fprintf` as done here:

https://github.com/ros2/rmw_connext/blob/aa5f47f6af79d2d4846dffdbf0394a6eaaaf80d6/rmw_connext_shared_cpp/src/node_names.cpp#L135-L137

That will default to printing to stderr, but allow someone to change where log messages are output.

jwillemsen: Created new issue #344 to fix all printf through a separate issue

sloretz: Works for me. Thanks for the issue and PR!

Issue #334, opened 2018-12-17
Shared memory transport still in use


hidmic:
## Bug report

**Required Info:**

**Required Info:**
- Operating System:
  - Ubuntu Bionic 18.04 
- Installation type:
  - Crystal pre-release binaries
- Version or commit hash:
  - 0.6.1
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

Perform the [QoS demo on lossy networks](https://index.ros.org/doc/ros2/Quality-of-Service/#add-network-traffic) using RTI Connext as your DDS vendor of choice.

#### Expected behavior
As the network degrades, topic subscription rates on a reliable topic degrade as well.
#### Actual behavior
No rate decrease is seen. Moreover, relevant nodes seem to be making use of shared memory.

```sh
michel@1bc3650db395:~$ ipcs -mp
------ Shared Memory Creator/Last-op PIDs --------
shmid      owner      cpid       lpid
65536      michel     13438      13495
98305      michel     13438      13495
229378     michel     13495      13438
262147     michel     13495      13495

michel@1bc3650db395:~$ ps ax
 PID TTY      STAT   TIME COMMAND
   1 pts/28   Ss     0:00 /bin/bash
13405 pts/28   S+     0:00 tmux
13407 ?        Ss     0:00 tmux
13408 pts/25   Ss     0:00 -bash
13437 pts/25   S+     0:00 /usr/bin/python3 /opt/ros/crystal/bin/ros2 run image_tools showimage
13438 pts/25   Rl+    0:56 /opt/ros/crystal/lib/image_tools/showimage
13490 pts/29   Ss     0:00 -bash
13494 pts/29   S+     0:00 /usr/bin/python3 /opt/ros/crystal/bin/ros2 run image_tools cam2image
13495 pts/29   Sl+    0:32 /opt/ros/crystal/lib/image_tools/cam2image
13519 pts/30   Ss     0:00 -bash
13532 pts/30   R+     0:00 ps ax
```
#### Additional information
RTI Connext API provides an additional [QoS policy setting](http://community.rti.com/rti-doc/510/ndds/doc/html/api_cpp/structDDS__TransportBuiltinQosPolicy.html) that is not being currently set. Not necessarily related though.

gonzodepedro:
I can reproduce this on Ubuntu 18.04 using crystal .deb packages. Also limiting traffic by using tc tool doesn't have any effect in the communication, it seems that shared memory is used when running locally.

[...]

PR #341, opened 2019-01-08
Force usage of UDPv4 even in local connections. Related #334

gonzodepedro:
DO NOT MERGE !!
This is a suggestion based on issue #334. It forces all traffic to be sent over UDPv4 and by doing that it prevents traffic to be sent over SharedMemory on localhost.
[Reference documentation](http://community.rti.com/rti-doc/510/ndds/doc/html/api_cpp/structDDS__TransportBuiltinQosPolicy.html)

### DIFF
--- a/rmw_connext_shared_cpp/src/node.cpp
+++ b/rmw_connext_shared_cpp/src/node.cpp
@@ -82,7 +82,7 @@ create_node(
   // So we set the limit to 1024, to accomodate the complete topic name with namespaces.
   participant_qos.resource_limits.contentfilter_property_max_length = 1024;
 
-  // forces local traffic to be sent over UDPv4,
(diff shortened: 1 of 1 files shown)


dirk-thomas:
While this would be the correct way to force Connext to use loopback instead of shared memory we don't want to go that route. Instead we should add a note to the [tutorial](https://index.ros.org/doc/ros2/Tutorials/Quality-of-Service/#add-network-traffic) about Connext and that it will use shared memory with the default configuration and therefore can't be affected by the traffic control rules.

@gonzodepedro Please create a PR for the documentation and then close this PR. Thanks!

gonzodepedro:
Note added on PR ros2/ros2_documentation/#97

PR #342, opened 2019-01-08
Fixed typo in comment

jwillemsen:
* rmw_connext_shared_cpp/src/node.cpp:


sloretz:
Thanks for the PR! Merging without full CI since it's an obviously safe change

PR #343, opened 2019-01-10
pass context to wait set and fini context

wjwwood:
Connects to ros2/rmw#163

### DIFF
--- a/rmw_connext_shared_cpp/src/wait_set.cpp
+++ b/rmw_connext_shared_cpp/src/wait_set.cpp
@@ -26,7 +26,7 @@ create_wait_set(
     context->implementation_identifier,
     implementation_identifier,
     // TODO(wjwwood): replace this with RMW_RET_INCORRECT_RMW_IMPLEMENTATION when refactored
-    return NULL);
+    return nullptr);
 
   rmw_wait_set_t * wait_set = rmw_wait_set_allocate();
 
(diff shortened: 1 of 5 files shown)


wjwwood:
I still need a review on this pr (and one other https://github.com/ros2/rmw_fastrtps/pull/252), if anyone has time.

rmw_connext_shared_cpp/src/wait_set.cpp
```diff
@@ -15,8 +15,19 @@
...
+  rmw_context_t * context,
+  size_t max_conditions)
 {
+  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context, NULL);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    init context,
+    context->implementation_identifier,
+    implementation_identifier,
+    // TODO(wjwwood): replace this with RMW_RET_INCORRECT_RMW_IMPLEMENTATION when refactored
+    return NULL);
```
clalancette: Ah, good, it looks like this method already returns `NULL` elsewhere, so callers already have to expect that.  I'll suggest `return nullptr` here to be consistent with the rest of the code, but it is minor.

wjwwood: I think my brain was just in C mode, I can update them.

PR #351, opened 2019-04-03
update patches for Connext 6, ignore pedantic warning

dirk-thomas:
I updated the trivial parts of the patches to apply to the code generated by Connext 6. The patch file `connext_static_serialized_dataPlugin.cxx.patch` contains the heavy lifting and still needs to be updated.

@Karsten1987 since you wrote these patches can you please take a look and add commits to this branch to make this work with Connext 6. The goal is to just make this work with the latest release of Connext - not change the way this is being done.

(It seems that the current logic applying the patches doesn't check if they apply cleanly and applies them anyway. So instead of failing when trying it with Connext 6 the patch files are applied even though the context doesn't match the actual files. Then it fails to compile the patched code since the content of the patched files doesn't make any sense. It would be nice if the logic which actually check that the patches apply cleanly - but that doesn't have to be par of this PR.)

### DIFF
--- a/rmw_connext_cpp/resources/patch_files/connext_static_serialized_data.h.patch
+++ b/rmw_connext_cpp/resources/patch_files/connext_static_serialized_data.h.patch
@@ -7,5 +7,5 @@
 -#include "ndds/ndds_cpp.h"
 +#include "rmw_connext_shared_cpp/ndds_include.hpp"
  #endif
+ #include "rti/xcdr/Interpreter.hpp"
  #else
- #include "ndds_standalone_type.h"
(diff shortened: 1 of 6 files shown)


dirk-thomas:
> It seems that the current logic applying the patches doesn't check if they apply cleanly and applies them anyway. So instead of failing when trying it with Connext 6 the patch files are applied even though the context doesn't match the actual files. Then it fails to compile the patched code since the content of the patched files doesn't make any sense. It would be nice if the logic which actually check that the patches apply cleanly - but that doesn't have to be par of this PR.

This has been addressed by #371.

Some updates necessary to support Connext 6 have landed in ros2/rosidl_typesupport_connext#37 and #374 which contains the conditional logic to choose between the major Connext versions.

I am closing this PR in favor of the following ticket describing the remaining work: #375.

PR #353, opened 2019-05-01
Stub out new allocation APIs.

mjcarroll:
Connects to ros2/rmw#159


rmw_connext_cpp/src/rmw_publisher.cpp
```diff
@@ -39,6 +39,27 @@
 
 extern "C"
 {
+rmw_ret_t
+rmw_init_publisher_allocation(
+  const rosidl_message_type_support_t * type_support,
+  const rosidl_message_bounds_t * message_bounds,
+  rmw_publisher_allocation_t * allocation)
+{
+  // Unused in current implementation.
```
clalancette: As with the other rmw implementations, I think we should probably return an error here and below.

PR #358, opened 2019-06-07
Implement get_actual_qos() for subscriptions

mm318:
Currently, publishers have the get_actual_qos() feature/function. It would make sense for subscriptions to have them, too.

### DIFF
--- a/rmw_connext_cpp/src/rmw_subscription.cpp
+++ b/rmw_connext_cpp/src/rmw_subscription.cpp
@@ -390,67 +390,7 @@ rmw_subscription_get_actual_qos(
     return RMW_RET_ERROR;
   }
 
-  if (!data_reader) {
-    RMW_SET_ERROR_MSG("subscription internal dds subscriber is invalid");
-    return RMW_RET_ERROR;
-  }
-
-  switch (dds_qos.history.kind) {
-    case DDS_KEEP_LAST_HISTORY_QOS:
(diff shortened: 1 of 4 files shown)


rmw_connext_cpp/src/rmw_subscription.cpp
```diff
@@ -365,6 +365,96 @@ rmw_subscription_count_matched_publishers(
...
+    return RMW_RET_ERROR;
+  }
+  DDS::DataReaderQos dds_qos;
+  DDS::ReturnCode_t status = data_reader->get_qos(dds_qos);
+  if (DDS::RETCODE_OK != status) {
+    RMW_SET_ERROR_MSG("subscription can't get data reader qos policies");
+    return RMW_RET_ERROR;
+  }
+
+  if (!data_reader) {
```
ivanpauno: Why is this check needed? I think the one in L382 is enough. I know that the same was done in:
https://github.com/ros2/rmw_connext/blob/8e9e08825c91b0a7abd9ef04028bbb515a3105f4/rmw_connext_cpp/src/rmw_publisher.cpp#L377-L393
But I don't think it has sense. If you agree, delete the duplicated check in the publisher function too

mm318: Done.

mm318:
Refactored to avoid duplicate code. Thanks!

ivanpauno:
CI: https://github.com/ros2/rmw/pull/177#issuecomment-501261185

PR #361, opened 2019-06-13
Add function for getting clients by node

jacobperron:
Connects to https://github.com/ros2/rmw/pull/179

### DIFF
--- a/rmw_connext_shared_cpp/src/node_info_and_types.cpp
+++ b/rmw_connext_shared_cpp/src/node_info_and_types.cpp
@@ -245,14 +245,16 @@ get_publisher_names_and_types_by_node(
   return copy_topics_names_and_types(topics, allocator, no_demangle, topic_names_and_types);
 }
 
+static
 rmw_ret_t
-get_service_names_and_types_by_node(
+__get_service_names_and_types_by_node(
   const char * implementation_identifier,
   const rmw_node_t * node,
   rcutils_allocator_t * allocator,
   const char * node_name,
   const char * node_namespace,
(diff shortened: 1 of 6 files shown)


rmw_connext_shared_cpp/src/node_info_and_types.cpp
```diff
@@ -291,4 +293,42 @@ get_service_names_and_types_by_node(
...
+  rmw_names_and_types_t * service_names_and_types)
+{
+  return __get_service_names_and_types_by_node(
+    implementation_identifier,
+    node,
+    allocator,
+    node_name,
+    node_namespace,
+    service_names_and_types,
+    "Reply");
```
mjcarroll: I know this is the last review of the set, but I just considered that maybe the "Reply" and "Request" should be some sort of static const, so that we don't have to update them if something changes?

jacobperron: Yeah, that's a good idea. What about `#define` them in [names_and_types_helpers.hpp](https://github.com/ros2/rmw_opensplice/blob/master/rmw_opensplice_cpp/src/names_and_types_helpers.hpp)?

jacobperron: If you're okay with it, I can do this refactoring in a follow-up set of PRs, since this can be done in all of the rmw implementations.

mjcarroll: I'm good with the followup.

PR #362, opened 2019-07-01
Filtering empty names with naive approach

CaptainTrunky:
Connects to ros2/ros2#489

A pull request for #489 basic implementation that allocates a temporary buffer for node names and filters out all empty names. It's not properly tested yet and additional unit tests are required. So next steps are:

- [ ] add tests (ros2/system_tests#374)

- [ ] add optimal implementation that does not require additional buffer and copying data

PS
had to reopen PR because rebasing all the 6+ month old changes  is way too painful to be productive.

Signed-off-by: captaintrunky <captaintrunky@gmail.com>

### DIFF
--- a/rmw_connext_shared_cpp/src/node_names.cpp
+++ b/rmw_connext_shared_cpp/src/node_names.cpp
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <iostream>
 #include <string>
 #include <vector>
 
@@ -48,38 +49,51 @@ get_node_names(
 
   DDS::DomainParticipant * participant = static_cast<ConnextNodeInfo *>(node->data)->participant;
   DDS::InstanceHandleSeq handles;
+
   if (participant->get_discovered_participants(handles) != DDS::RETCODE_OK) {
     RMW_SET_ERROR_MSG("unable to fetch discovered participants.");
     return RMW_RET_ERROR;
   }
-  auto length = handles.length() + 1;  // add yourself
-  rcutils_allocator_t allocator = rcutils_get_default_allocator();
-  rcutils_ret_t rcutils_ret = rcutils_string_array_init(node_names, length, &allocator);
-  if (rcutils_ret != RCUTILS_RET_OK) {
-    RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
-    return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
+
+  DDS::DomainParticipantQos participant_qos;
+  DDS::ReturnCode_t status = participant->get_qos(participant_qos);
+
+  if (status != DDS::RETCODE_OK) {
+    RMW_SET_ERROR_MSG("failed to get default participant qos");
+    return RMW_RET_ERROR;
   }
-  rcutils_ret = rcutils_string_array_init(node_namespaces, length, &allocator);
+
+  auto length = handles.length() + 1;  // add yourself
+
+  rcutils_allocator_t allocator = rcutils_get_default_allocator();
+  
+  // Pre-allocate temporary buffer for all node names & namespaces
+  // Nodes that are not created by ROS 2 could provide empty names
+  // Such names should not be returned
+  rcutils_string_array_t tmp_names_list = rcutils_get_zero_initialized_string_array();
+  rcutils_ret_t rcutils_ret = rcutils_string_array_init(&tmp_names_list, length, &allocator);
+
   if (rcutils_ret != RCUTILS_RET_OK) {
     RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
     return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
   }
 
-  DDS::DomainParticipantQos participant_qos;
-  DDS::ReturnCode_t status = participant->get_qos(participant_qos);
-  if (status != DDS::RETCODE_OK) {
-    RMW_SET_ERROR_MSG("failed to get default participant qos");
-    return RMW_RET_ERROR;
+  rcutils_string_array_t tmp_namespaces_list = rcutils_get_zero_initialized_string_array();
+  rcutils_ret = rcutils_string_array_init(&tmp_namespaces_list, length, &allocator);
+
+  if (rcutils_ret != RCUTILS_RET_OK) {
+    RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
+    return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
   }
-  node_names->data[0] = rcutils_strdup(participant_qos.participant_name.name, allocator);
-  if (!node_names->data[0]) {
+
+  int named_nodes_num = 0;
+  int named_namespaces_num = 0;
+ 
+  // add yourself
+  tmp_names_list.data[0] = rcutils_strdup(participant_qos.participant_name.name, allocator);
+  if (!tmp_names_list.data[0]) {
     RMW_SET_ERROR_MSG("could not allocate memory for node name");
-    return RMW_RET_BAD_ALLOC;
-  }
-  node_namespaces->data[0] = rcutils_strdup(node->namespace_, allocator);
-  if (!node_names->data[0]) {
-    RMW_SET_ERROR_MSG("could not allocate memory for node namespace");
-    return RMW_RET_BAD_ALLOC;
+    goto fail;
   }
 
   for (auto i = 1; i < length; ++i) {
@@ -90,16 +104,20 @@ get_node_names(
     if (DDS::RETCODE_OK == dds_ret) {
       auto data = static_cast<unsigned char *>(pbtd.user_data.value.get_contiguous_buffer());
       std::vector<uint8_t> kv(data, data + pbtd.user_data.value.length());
+
       auto map = rmw::impl::cpp::parse_key_value(kv);
+
       auto name_found = map.find("name");
       auto ns_found = map.find("namespace");
 
       if (name_found != map.end()) {
         name = std::string(name_found->second.begin(), name_found->second.end());
       }
-      if (name_found != map.end()) {
+
+      if (ns_found != map.end()) {
         namespace_ = std::string(ns_found->second.begin(), ns_found->second.end());
       }
+      
       if (name.empty()) {
         // use participant name if no name was found in the user data
         if (pbtd.participant_name.name) {
@@ -107,27 +125,79 @@ get_node_names(
         }
       }
     }
+
+    // ignore discovered participants without a name
     if (name.empty()) {
-      // ignore discovered participants without a name
-      node_names->data[i] = nullptr;
-      node_namespaces->data[i] = nullptr;
+      tmp_names_list.data[i] = nullptr;
+      tmp_namespaces_list.data[i] = nullptr;
       continue;
     }
 
-    node_names->data[i] = rcutils_strdup(name.c_str(), allocator);
-    if (!node_names->data[i]) {
+    tmp_names_list.data[named_nodes_num] = rcutils_strdup(name.c_str(), allocator);
+    if (!tmp_names_list.data[named_nodes_num]) {
       RMW_SET_ERROR_MSG("could not allocate memory for node name");
       goto fail;
     }
 
-    node_namespaces->data[i] = rcutils_strdup(namespace_.c_str(), allocator);
-    if (!node_namespaces->data[i]) {
+    ++named_nodes_num;
+    
+    if (namespace_.empty()) {
+      tmp_namespaces_list.data[i] = nullptr;
+      continue;
+    }
+    
+    tmp_namespaces_list.data[named_namespaces_num] = rcutils_strdup(namespace_.c_str(), allocator);
+    if (!tmp_namespaces_list.data[named_namespaces_num]) {
       RMW_SET_ERROR_MSG("could not allocate memory for node namespace");
       goto fail;
     }
+    
+    ++named_namespaces_num;
   }
 
+  // prepare actual output without empty names
+  // TODO: implement functionality that allocates just enough memory for the output buffer,
+  // so using additional buffer would be unnessecary 
+  rcutils_ret = rcutils_string_array_init(node_names, named_nodes_num, &allocator);
+  if (rcutils_ret != RCUTILS_RET_OK) {
+    RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
+    return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
+  }
+
+  for (auto i = 0; i < named_nodes_num; ++i) {
+    node_names->data[i] = rcutils_strdup(tmp_names_list.data[i], allocator);
+    tmp_names_list.data[i] = nullptr;
+  }
+
+  for (auto i = 0; i < named_nodes_num; ++i) {
+    std::cerr << tmp_names_list.data[i] << '\n';
+  }
+  rcutils_ret = rcutils_string_array_fini(&tmp_names_list);
+  if (rcutils_ret != RCUTILS_RET_OK) {
+    RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
+    return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
+  }
+  
+  rcutils_ret = rcutils_string_array_init(node_namespaces, named_namespaces_num, &allocator);
+
+  if (rcutils_ret != RCUTILS_RET_OK) {
+    RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
+    return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
+  }
+  
+  for (auto i = 0; i < named_namespaces_num; ++i) {
+    node_namespaces->data[i] = rcutils_strdup(tmp_namespaces_list.data[i], allocator);
+    tmp_namespaces_list.data[i] = nullptr;
+  }
+  
+  rcutils_ret = rcutils_string_array_fini(&tmp_namespaces_list);
+  if (rcutils_ret != RCUTILS_RET_OK) {
+    RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
+    return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
+  }
+  
   return RMW_RET_OK;
+
 fail:
   if (node_names) {
     rcutils_ret = rcutils_string_array_fini(node_names);
@@ -138,14 +208,35 @@ fail:
       rcutils_reset_error();
     }
   }
+
   if (node_namespaces) {
     rcutils_ret = rcutils_string_array_fini(node_namespaces);
     if (rcutils_ret != RCUTILS_RET_OK) {
       RCUTILS_LOG_ERROR_NAMED(
         "rmw_connext_cpp",
-        "failed to cleanup during error handling: %s", rcutils_get_error_string().str);
+        "failed to cleanup during error handling: %s", rcutils_get_error_string().str
+      );
       rcutils_reset_error();
     }
   }
+  
+  rcutils_ret = rcutils_string_array_fini(&tmp_names_list);
+  if (rcutils_ret != RCUTILS_RET_OK) {
+    RCUTILS_LOG_ERROR_NAMED(
+      "rmw_connext_cpp",
+      "failed to cleanup during error handling: %s", rcutils_get_error_string().str
+    );
+    rcutils_reset_error();
+  }
+ 
+  rcutils_ret = rcutils_string_array_fini(&tmp_namespaces_list);
+  if (rcutils_ret != RCUTILS_RET_OK) {
+    RCUTILS_LOG_ERROR_NAMED(
+      "rmw_connext_cpp",
+      "failed to cleanup during error handling: %s", rcutils_get_error_string().str
+    );
+    rcutils_reset_error();
+  }
+  
   return RMW_RET_BAD_ALLOC;
 }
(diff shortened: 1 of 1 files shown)


CaptainTrunky:
Finally I've finished merging with current master.

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -96,27 +106,28 @@ get_node_names(
 
       if (name_found != map.end()) {
         name = std::string(name_found->second.begin(), name_found->second.end());
-      }
-      if (name_found != map.end()) {
```
jacobperron: Looks like this was previously a bug. Instead of removing the condition, I think it should be:

```c++
if (ns_found != map.end()) {
```

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -96,27 +106,28 @@ get_node_names(
...
         namespace_ = std::string(ns_found->second.begin(), ns_found->second.end());
       }
+
+      // ignore discovered participants without a name
```
jacobperron: I think this comment isn't necessary. The block below is not ignoring discovered participants, it's using the participant name instead of the name from user data in the case that is is empty.

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -96,27 +106,28 @@ get_node_names(
...
-      node_names->data[i] = nullptr;
+      tmp_names_list.data[i] = nullptr;
       node_namespaces->data[i] = nullptr;
       continue;
     }
 
-    node_names->data[i] = rcutils_strdup(name.c_str(), allocator);
-    if (!node_names->data[i]) {
-      RMW_SET_ERROR_MSG("could not allocate memory for node name");
+    tmp_names_list.data[named_nodes_num] = rcutils_strdup(name.c_str(), allocator);
```
jacobperron: I think this should use `i`, as we're using it on the following line.
```suggestion
    tmp_names_list.data[i] = rcutils_strdup(name.c_str(), allocator);
```

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -96,27 +106,28 @@ get_node_names(
...
+    if (!tmp_names_list.data[i]) {
+      RMW_SET_ERROR_MSG("could not allocate memory for node namespace");
```
jacobperron:
```suggestion
      RMW_SET_ERROR_MSG("could not allocate memory for node name");
```

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -125,26 +136,63 @@ get_node_names(
...
+  // TODO: implement functionality that allocates just enough memory for the output buffer,
+  // so using additional buffer would be unnessecary 
+  rcutils_ret = rcutils_string_array_init(node_names, named_nodes_num, &allocator);
+  if (rcutils_ret != RCUTILS_RET_OK) {
+    RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
+    return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
+  }
+
+  for (auto i = 0; i < named_nodes_num; ++i) {
+    node_names->data[i] = rcutils_strdup(tmp_names_list.data[i], allocator);
```
jacobperron: This logic looks incorrect. It's possible there is an empty name in the middle of `tmp_names_list`, in which case we're copying `nullptr` (and missing a potential valid name at the end of the list).

The for-loop should loop up to `length` and then inside we should guard against `nullptr`:

```c++
size_t j = 0;
for (auto i = 0; i < length; ++i) {
  if (tmp_names_list.data[i]) {
    node_names->data[j] = rcutils_strdup(tmp_names_list.data[i], allocator);
  }
  ++j;
}
```

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -125,26 +136,63 @@ get_node_names(
...
+    tmp_names_list.data[i] = nullptr;
   }
 
+  // prepare actual output without empty names
+  // TODO: implement functionality that allocates just enough memory for the output buffer,
+  // so using additional buffer would be unnessecary 
+  rcutils_ret = rcutils_string_array_init(node_names, named_nodes_num, &allocator);
```
jacobperron: This block and the next are duplicated.

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -125,26 +136,63 @@ get_node_names(
...
+
 fail:
   if (node_names) {
     rcutils_ret = rcutils_string_array_fini(node_names);
     if (rcutils_ret != RCUTILS_RET_OK) {
-      RCUTILS_LOG_ERROR_NAMED(
-        "rmw_connext_cpp",
-        "failed to cleanup during error handling: %s", rcutils_get_error_string().str);
-      rcutils_reset_error();
+      RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
```
jacobperron: In all cases where we `goto fail`, the error message is already set. You can undo this change.

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -125,26 +136,63 @@ get_node_names(
...
+      return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
     }
   }
+  rcutils_ret = rcutils_string_array_fini(&tmp_names_list);
+  if (rcutils_ret != RCUTILS_RET_OK) {
+    RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
+    return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
```
jacobperron: Instead of setting an error message and returning, log an error and reset the error:

```c++
RCUTILS_LOG_ERROR_NAMED(
  "rmw_connext_cpp",
  "failed to cleanup during error handling: %s", rcutils_get_error_string().str);
rcutils_reset_error();
```

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -125,26 +136,63 @@ get_node_names(
...
+  }
   if (node_namespaces) {
     rcutils_ret = rcutils_string_array_fini(node_namespaces);
     if (rcutils_ret != RCUTILS_RET_OK) {
-      RCUTILS_LOG_ERROR_NAMED(
-        "rmw_connext_cpp",
-        "failed to cleanup during error handling: %s", rcutils_get_error_string().str);
-      rcutils_reset_error();
+      RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
+      return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
```
jacobperron: Undo this change.

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -45,41 +45,51 @@ get_node_names(
...
+  rcutils_ret_t rcutils_ret = rcutils_string_array_init(node_namespaces, length, &allocator);
+
   if (rcutils_ret != RCUTILS_RET_OK) {
     RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
     return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
   }
-  rcutils_ret = rcutils_string_array_init(node_namespaces, length, &allocator);
+
+  // Pre-allocate temporary buffer for all node names
+  // Nodes that are not created by ROS2 could provide empty names
```
jacobperron: nitpick: "ROS 2" (or just "ROS").

nuclearsandwich:
@CaptainTrunky do you have plans to resume iterating on this PR? Thanks.

jacobperron:
@CaptainTrunky Friendly ping :slightly_smiling_face: Let us know if you'd like to keep working on this.

CaptainTrunky:
Sorry, I've been busy with my job. I'll try to resolve current issues within next two weeks.

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: I guess that that's not right. My idea was to overallocate `tmp_names_list`, so it might be identical to `node_names`. Next, I add node names to tmp_names_list only if a name is not empty. Thus, `named_nodes_num` tracks number of non-empty names so far. After traversing the whole list of names, `tmp_names_list` contains valid names and (possibly) nullptrs at the end. We can safely truncate them later. Is this logic applicable?

jacobperron: Okay, makes sense. We should do the same logic for the namespaces as well, no?

rmw_connext_shared_cpp/src/node_names.cpp
jacobperron: I overlooked the logic above (https://github.com/ros2/rmw_connext/pull/362/files#r327540360), this loop looks okay.

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: I think so. This function treats names and namespaces in almost identical way.

dirk-thomas:
@CaptainTrunky Another ping

CaptainTrunky:
I've added support for namespaces. Plan to to test it on Wednesday and push final version.

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: Done

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: Done

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: Done

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: Done

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: Done

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: Done

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: Done

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: Done

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: Done

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -90,44 +104,100 @@ get_node_names(
...
+    ++named_namespaces_num;
   }
 
+  // prepare actual output without empty names
+  // TODO: implement functionality that allocates just enough memory for the output buffer,
+  // so using additional buffer would be unnessecary 
+  rcutils_ret = rcutils_string_array_init(node_names, named_nodes_num, &allocator);
+  if (rcutils_ret != RCUTILS_RET_OK) {
+    RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
+    return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
```
jacobperron: We should `goto fail;` here so we can cleanup.

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -90,44 +104,100 @@ get_node_names(
...
-      node_names->data[i] = nullptr;
-      node_namespaces->data[i] = nullptr;
+      tmp_names_list.data[i] = nullptr;
+      tmp_namespaces_list.data[i] = nullptr;
       continue;
     }
 
-    node_names->data[i] = rcutils_strdup(name.c_str(), allocator);
-    if (!node_names->data[i]) {
+    tmp_names_list.data[named_nodes_num] = rcutils_strdup(name.c_str(), allocator);
```
jacobperron: This will overwrite the first element of `tmp_names_list` that was set at L93. I think `named_nodes_num` should be initialized to `1`. Same for `named_namespaces_num`.

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -90,44 +104,100 @@ get_node_names(
...
+    RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
+    return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
+  }
+
+  for (auto i = 0; i < named_nodes_num; ++i) {
+    node_names->data[i] = rcutils_strdup(tmp_names_list.data[i], allocator);
+    tmp_names_list.data[i] = nullptr;
+  }
+
+  for (auto i = 0; i < named_nodes_num; ++i) {
```
jacobperron: I guess this loop is leftover from debugging and can be removed.

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <iostream>
```
jacobperron: I think this can be removed.

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -90,44 +104,100 @@ get_node_names(
...
       RMW_SET_ERROR_MSG("could not allocate memory for node name");
       goto fail;
     }
 
-    node_namespaces->data[i] = rcutils_strdup(namespace_.c_str(), allocator);
-    if (!node_namespaces->data[i]) {
+    ++named_nodes_num;
+    
+    if (namespace_.empty()) {
+      tmp_namespaces_list.data[i] = nullptr;
```
jacobperron: This is concerning, since we expect that the output names and namespaces lists will be the same size representing a one-to-one mapping for nodes ([docs](https://github.com/ros2/rmw/blob/813b94ddd5650444b312304ad156f3b5d9f05e39/rmw/include/rmw/rmw.h#L1015-L1018)). It is possible that if we have empty namespaces, but not empty names, that we will end up with lists of different sizes and lose the mapping. In other words, we want to ensure that `named_nodes_num == named_namespaces_num`.

I think that if the namespace is empty, we can let it be empty and not filter it like we do with the node names. We can probably just remove this block.

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -90,44 +104,100 @@ get_node_names(
...
+    tmp_names_list.data[i] = nullptr;
+  }
+
+  for (auto i = 0; i < named_nodes_num; ++i) {
+    std::cerr << tmp_names_list.data[i] << '\n';
+  }
+  rcutils_ret = rcutils_string_array_fini(&tmp_names_list);
+  if (rcutils_ret != RCUTILS_RET_OK) {
+    RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
+    return rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
```
jacobperron: Be careful returning too early, we are leaking memory since `tmp_namespaces_list` isn't being finalized.
I suggest refactoring so that we only return once at the end of the function. We can rename the `fail` label to `cleanup`, and whenever there is an error, we can set an error message, set a return code, and `goto cleanup`.

E.g. 

```c
rmw_ret_t final_ret = RMW_RET_OK;  // put this near the top after the initial error handling (e.g. L65)

// ...

if (rcutils_ret_ != RCUTILS_RET_OK) {
  RMW_SET_ERROR_MSG(rcutils_get_error_string().str);
  final_ret = rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
  goto cleanup;
}

// ...

cleanup:
// ... cleanup logic
return final_ret;
```

rmw_connext_shared_cpp/src/node_names.cpp
jacobperron: In fact, we might remove the variable `named_namespaces_num` altogether since it should always be the same as `named_nodes_num`.

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: Done

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: Done

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: Done

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: Done

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: Done

rmw_connext_shared_cpp/src/node_names.cpp
CaptainTrunky: Done

jacobperron:
Sorry for the delay in getting back to this. I've rebased on master and added a small fix (70e6973) and resolved linter errors (68f59bb).

Running CI for all platforms to see where we stand:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8737/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4566/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7139/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8651/)

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -48,38 +48,56 @@ get_node_names(
...
-  if (!node_names->data[0]) {
-    RMW_SET_ERROR_MSG("could not allocate memory for node name");
-    return RMW_RET_BAD_ALLOC;
-  }
-  node_namespaces->data[0] = rcutils_strdup(node->namespace_, allocator);
-  if (!node_names->data[0]) {
-    RMW_SET_ERROR_MSG("could not allocate memory for node namespace");
-    return RMW_RET_BAD_ALLOC;
+  // add yourself
+  tmp_names_list.data[0] = rcutils_strdup(participant_qos.participant_name.name, allocator);
```
dirk-thomas: This only populates the `name` bub not the `namespace` of the current node. As a result a test in `rcl` is failing since this PR was merged.

The CI builds ran for this change also didn't cover `rcl` as far as I can see. Please address this regression as soon as possible.

jacobperron: My mistake; I only ran test for this package and test_communication. I'll fix this.

jacobperron: See https://github.com/ros2/rmw_connext/pull/384 for a fix.

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -48,38 +48,56 @@ get_node_names(
...
   rcutils_allocator_t allocator = rcutils_get_default_allocator();
-  rcutils_ret_t rcutils_ret = rcutils_string_array_init(node_names, length, &allocator);
+
+  // Pre-allocate temporary buffer for all node names & namespaces
+  // Nodes that are not created by ROS 2 could provide empty names
+  // Such names should not be returned
+  rcutils_string_array_t tmp_names_list = rcutils_get_zero_initialized_string_array();
+  rcutils_string_array_t tmp_namespaces_list = rcutils_get_zero_initialized_string_array();
+
+  int named_nodes_num = 1;
```
ivanpauno: why this starts from 1?

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -90,45 +108,93 @@ get_node_names(
...
+
+  if (rcutils_ret != RCUTILS_RET_OK) {
+    RMW_SET_ERROR_MSG("could not allocate memory for node_namespaces output");
+    final_ret = rmw_convert_rcutils_ret_to_rmw_ret(rcutils_ret);
+    goto cleanup;
+  }
+
+  for (auto i = 0; i < named_nodes_num; ++i) {
+    node_names->data[i] = rcutils_strdup(tmp_names_list.data[i], allocator);
+    tmp_names_list.data[i] = nullptr;
```
ivanpauno: isn't this leaking memory?

rmw_connext_shared_cpp/src/node_names.cpp
jacobperron: The first element is set to the current node name at L96. I guess for clarity this could have started at zero and then been incremented after L96.

rmw_connext_shared_cpp/src/node_names.cpp
jacobperron: Yes, I suppose it does. Good catch! I don't think this line or the other at L178 are necessary.

See https://github.com/ros2/rmw_connext/pull/395 for a fix.

rmw_connext_shared_cpp/src/node_names.cpp
ivanpauno: Yes, didn't realize.
The logic is a bit confusing.

PR #370, opened 2019-10-11
suppress invalid syntax warning for macro call

Karsten1987:
use the inline suppressing function of cppcheck to disable the `syntaxError` warnings shown here:
https://ci.ros2.org/view/nightly/job/nightly_win_rel/1338/testReport/junit/(root)/projectroot/cppcheck_2/

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/8347/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6803/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8249/)

Signed-off-by: Karsten Knese <karsten@openrobotics.org>

### DIFF
--- a/rmw_connext_cpp/src/rmw_client.cpp
+++ b/rmw_connext_cpp/src/rmw_client.cpp
@@ -179,6 +179,7 @@ rmw_create_client(
     goto fail;
   }
   // Use a placement new to construct the ConnextStaticClientInfo in the preallocated buffer.
+  // cppcheck-suppress syntaxError
   RMW_TRY_PLACEMENT_NEW(client_info, buf, goto fail, ConnextStaticClientInfo, )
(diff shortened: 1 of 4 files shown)


Karsten1987:
alternatively we could change the macro accordingly so that we have essentially two macros. 
We could rename the existing macros into something like `TRY_PLACEMENT_NEW_WITH_ARGS` where the VA_ARGS are passed explicitly to the constructor and a second one which is just called `TRY_PLACEMENT_NEW`, which doesn't have the `...` support and thus avoids the trailing `,`.

Karsten1987:
for completeness, rerunning linux ci:
[Build Status](https://ci.ros2.org/job/ci_linux/8376/)

PR #371, opened 2019-10-15
assert that unmodified lines in the patch match the input

dirk-thomas:
Otherwise an incorrectly applied patch file doesn't cause an error during patching (and we can only hope for a compiler error later).

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8357/) (assuming these test failures are unrelated)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6818/) (failing on master with the same problems)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8267/) (assuming these test failures are unrelated)

### DIFF
--- a/rmw_connext_cpp/bin/apply-patch.py
+++ b/rmw_connext_cpp/bin/apply-patch.py
@@ -54,6 +54,11 @@ def apply_patch(s, patch):
                 line = p[i]
                 i += 1
             if len(line) > 0:
+                if line[0] == ' ':
+                    # ensure that any unmodified line is the same
+                    # in the input file and in the patch file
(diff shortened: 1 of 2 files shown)


ivanpauno:
> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/8357/) (assuming these test failures are unrelated)
> * macOS [Build Status](http://ci.ros2.org/job/ci_osx/6818/) (failing on master with the same problems)
> * Windows [Build Status](http://ci.ros2.org/job/ci_windows/8267/) (assuming these test failures are unrelated)

It's not clear to me if the failures are unrelated or not.

PR #378, opened 2019-10-22
CMAKE_SOURCE_DIR -> CMAKE_CURRENT_SOURCE_DIR

rotu:
These will differ if this project is built as a sub-project of another cmake project, and in that case, CMAKE_CURRENT_SOURCE_DIR will point to the right place.

Signed-off-by: Dan Rose <dan@digilabs.io>


dirk-thomas:
Can you please add a description why you made the change.

rotu:
@dirk-thomas Done

dirk-thomas:
Thanks.

Issue #381, opened 2019-12-20
Refactor code to use one Participant per Context


ivanpauno:
#### Feature description

See https://github.com/ros2/rmw/issues/183.

clalancette:
Closing, since with Foxy now being [End-of-Life](https://discourse.ros.org/t/new-packages-for-foxy-fitzroy-2023-06-20/32039), this repository is no longer used.

PR #382, opened 2019-12-20
Groundwork for one participant per context refactor

ivanpauno:
Groundwork for https://github.com/ros2/rmw_connext/issues/381.

Just moving code around, shouldn't change behavior.
This allows to:
- Create a rmw_publisher_t from a participant.
- Create a rmw_subscription_t from a participant.

This refactor can be merged right now, avoiding hard rebases later.


rmw_connext_cpp/src/publisher.cpp
```diff
@@ -0,0 +1,376 @@
...
+  DDS::Publisher * dds_publisher = nullptr;
+  DDS::DataWriter * topic_writer = nullptr;
+  DDS::Topic * topic = nullptr;
+  DDS::TopicDescription * topic_description = nullptr;
+  void * info_buf = nullptr;
+  void * listener_buf = nullptr;
+  ConnextPublisherListener * publisher_listener = nullptr;
+  ConnextStaticPublisherInfo * publisher_info = nullptr;
+  rmw_publisher_t * publisher = nullptr;
+  std::string mangled_name = "";
```
hidmic: @ivanpauno nit: no need to initialize with an empty string.

rmw_connext_cpp/src/rmw_service.cpp
```diff
@@ -56,7 +56,7 @@ rmw_create_service(
     return nullptr;
   }
 
-  auto node_info = static_cast<ConnextNodeInfo *>(node->data);
+  auto node_info = static_cast<ConnextParticipantInfo *>(node->data);
```
hidmic: @ivanpauno here and elsewhere, `ConnextNodeInfo` became `ConnextParticipantInfo`, but I don't see the `struct` defined and `node_info` is probably not the right name anymore.

rmw_connext_cpp/src/publisher.cpp
ivanpauno: I actually copy-pasted that, but ok.

rmw_connext_cpp/src/rmw_service.cpp
ivanpauno: The struct is defined [here](https://github.com/ros2/rmw_connext/blob/4e2e037593cf27fa1e1201bb333fb2e7680a1971/rmw_connext_shared_cpp/include/rmw_connext_shared_cpp/types.hpp#L129).

And about renaming `node_info` to `participant_info`, I did that in https://github.com/ros2/rmw_connext/pull/382/commits/4e2e037593cf27fa1e1201bb333fb2e7680a1971.

rmw_connext_cpp/src/publisher.cpp
ivanpauno: Addressed in https://github.com/ros2/rmw_connext/pull/382/commits/4e2e037593cf27fa1e1201bb333fb2e7680a1971.

ivanpauno:
Closing as this is not going to be continued (considering that https://github.com/rticommunity/rmw_connextdds already implemented this new approach).

PR #392, opened 2020-02-28
API changes to sync with one Participant per Context change in rmw_fastrtps

ivanpauno:
Needed since ros2/rmw#189.


rmw_connext_shared_cpp/src/node.cpp
```diff
@@ -78,7 +73,8 @@ create_node(
   // since the participant name is not part of the DDS spec
   // the node name is also set in the user_data
   size_t length = strlen(name) + strlen("name=;") +
-    strlen(namespace_) + strlen("namespace=;") + 1;
+    strlen(namespace_) + strlen("namespace=;") +
+    strlen(context->options.security_context) + strlen("securitycontext=;") + 1;
```
hidmic: @ivanpauno I saw this being called `contextname` elsewhere. Any reason not to standardize?

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -127,6 +153,11 @@ get_node_names(
       if (ns_found != map.end()) {
         namespace_ = std::string(ns_found->second.begin(), ns_found->second.end());
       }
+
+      if (security_context_found != map.end()) {
+        security_context = std::string(
+          security_context_found->second.begin(), security_context_found->second.end());
```
hidmic: @ivanpauno this is a peculiar way of copying an `std::string` and I suspect less efficient than a plain copy-assignment.

rmw_connext_cpp/src/rmw_init.cpp
```diff
@@ -51,20 +57,38 @@ rmw_init_options_copy(const rmw_init_options_t * src, rmw_init_options_t * dst)
     RMW_SET_ERROR_MSG("expected zero-initialized dst");
     return RMW_RET_INVALID_ARGUMENT;
   }
+  const rcutils_allocator_t * allocator = &src->allocator;
```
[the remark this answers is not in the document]
ivanpauno: I can use a reference instead of a pointer.
About constness, we've some functions that take a [const pointer](https://github.com/ros2/rcutils/blob/af3e14543ffa7e8d4d9d38f9c242548f76accd89/include/rcutils/types/array_list.h#L126), there are others taking it [by value](https://github.com/ros2/rcutils/blob/af3e14543ffa7e8d4d9d38f9c242548f76accd89/include/rcutils/types/string_map.h#L99), and others taking a [mutable pointer](https://github.com/ros2/rmw/blob/f76511333709a6f5d1e4543129ec73f2cd2b4f06/rmw/include/rmw/names_and_types.h#L72).

In the first and last case, you can do:
```cpp
allocator->allocate(size, allocator->state);
```
In the second case, you can replace `->` with `.`.

Why does [this](https://github.com/ros2/rcutils/blob/af3e14543ffa7e8d4d9d38f9c242548f76accd89/src/array_list.c#L71) work? I will have to check, but I think that C structs constness is not propagated to its members. So, it doesn't matter that [allocate](https://github.com/ros2/rcutils/blob/af3e14543ffa7e8d4d9d38f9c242548f76accd89/include/rcutils/allocator.h#L49) takes a mutable pointer to `state`.

rmw_connext_shared_cpp/src/node.cpp
ivanpauno: It was supposed to be `securitycontext` everywhere. I forgot to update some places.

rmw_connext_shared_cpp/src/node_names.cpp
ivanpauno: See https://github.com/ros2/rmw/blob/f76511333709a6f5d1e4543129ec73f2cd2b4f06/rmw/include/rmw/impl/cpp/key_value.hpp#L31.

rmw_connext_cpp/src/rmw_init.cpp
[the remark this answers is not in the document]
hidmic: > I can use a reference instead of a pointer.

I was mostly interested in the rationale behind explicitly introducing an indirection. 

> About constness, we've some functions that take a const pointer, there are others taking it by value, and others taking a mutable pointer.

I'm aware of how allocators are passed in all forms and colors here and there. That's something we can discuss in a follow-up issue though.

> Why does this work? I will have to check, but I think that C structs constness is not propagated to its members. So, it doesn't matter that allocate takes a mutable pointer to state.

A `const` qualified `struct` in C/C++ implies that none of its members can be mutated. In the case you mention, you're actually not mutating `state`, the pointer. AFAIK cv-qualifiers do not propagate through indirections. That's why a `const std::shared_ptr<SomeObject>` will happily let you mutate `SomeObject`, whereas `const std::shared_ptr<const SomeObject>` will likely get you what you want.

rmw_connext_shared_cpp/src/node_names.cpp
hidmic: That's a peculiar API. I wonder why a sequence of bytes. Is it to support other encodings in the future? cc @dirk-thomas

rmw_connext_shared_cpp/src/node_names.cpp
```diff
@@ -169,9 +210,22 @@ get_node_names(
...
+      goto cleanup;
+    }
+  }
+
   for (auto i = 0; i < named_nodes_num; ++i) {
     node_names->data[i] = rcutils_strdup(tmp_names_list.data[i], allocator);
     node_namespaces->data[i] = rcutils_strdup(tmp_namespaces_list.data[i], allocator);
+    if (security_contexts) {
+      security_contexts->data[i] = rcutils_strdup(tmp_security_contexts_list.data[i], allocator);
+    }
```
hidmic: @ivanpauno meta: this predates your PR, but why does this code assume that `rcutils_strdup` will succeed? Also, simply moving strings in `tmp` arrays would be much more efficient than this (de)allocation dance.

ivanpauno: I completely agree.
I will open an issue for this, because it's out of the scope of this PR IMO.

rmw_connext_shared_cpp/src/node.cpp
ivanpauno: See https://github.com/ros2/rmw_connext/pull/392/commits/3d53fafa80a7da972c4d6ed48e04ffdac45cc866.

rmw_connext_cpp/src/rmw_init.cpp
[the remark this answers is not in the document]
ivanpauno: > I was mostly interested in the rationale behind explicitly introducing an indirection.

Ok, I can use a reference if you want.

> I'm aware of how allocators are passed in all forms and colors here and there. That's something we can discuss in a follow-up issue though.

:+1:

> A const qualified struct in C/C++ implies that none of its members can be mutated. In the case you mention, you're actually not mutating state, the pointer. AFAIK cv-qualifiers do not propagate through indirections. That's why a const std::shared_ptr<SomeObject> will happily let you mutate SomeObject, whereas const std::shared_ptr<const SomeObject> will likely get you what you want.

Yes, exactly. I misunderstood your first comment and derailed a bit.

rmw_connext_shared_cpp/src/node.cpp
```diff
@@ -72,16 +73,18 @@ create_node(
...
+    "name=%s;namespace=%s;securitycontext=%s;",
+    name, namespace_, context->options.security_context) + 1;
   bool success = participant_qos.user_data.value.length(static_cast<DDS::Long>(length));
   if (!success) {
     RMW_SET_ERROR_MSG("failed to resize participant user_data");
     return NULL;
   }
 
-  int written = snprintf(
+  int written = std::snprintf(
```
hidmic: @ivanpauno uh, why the added prefix?

hidmic: Also, I just remembered that `rcutils` has a portable [`snprintf`](https://github.com/ros2/rcutils/blob/af3e14543ffa7e8d4d9d38f9c242548f76accd89/include/rcutils/snprintf.h#L29-L58) flavor.

ivanpauno: > @ivanpauno uh, why the added prefix?

it's not really needed. I can delete it again.

> Also, I just remembered that rcutils has a portable snprintf flavor.

This seems to be portable, as it's currently working fine.
Do you know the rationale of having a `rcutils` flavor?

hidmic: Not surprisingly, there's something peculiar about using `snprintf` in Windows. But @wjwwood may have deeper insight.

ivanpauno: I mean, this code is passing CI on Windows with `master`, which does the same thing.
If CI is green, I wouldn't change anything else.

hidmic: Fair enough. I won't block on it, though now I'm curious about `rcutils_snprintf`.

wjwwood: Maybe the std:: avoids it but on Windows it will usually generate a compiler warning if you try to use snprintf directly. Or at least that’s my recollection.

PR #394, opened 2020-03-06
Increase the max_objects_per_thread.

clalancette:
The default is 1024, which allows about 11 participants.
According to the documentation at
https://community.rti.com/kb/how-many-domain-participants-can-be-created-single-address-space ,
2048 allows about 30.  Based on empirical evidence, we increase
to 8192, which should allow something like 60 - 70 participants.

### DIFF
--- a/rmw_connext_shared_cpp/src/init.cpp
+++ b/rmw_connext_shared_cpp/src/init.cpp
@@ -25,5 +25,10 @@ init()
     RMW_SET_ERROR_MSG("failed to get participant factory");
     return RMW_RET_ERROR;
   }
+
+  DDS::DomainParticipantFactoryQos factory_qos;
+  dpf_->get_qos(factory_qos);
+  factory_qos.resource_limits.max_objects_per_thread = 8192;
+  dpf_->set_qos(factory_qos);
   return RMW_RET_OK;
 }


clalancette:
Tested with https://github.com/clalancette/many_nodes/blob/7c0c44874be4421ce39e69746554284713a6c209/launch/many_nodes-launch.py ; there are comments in there, but before this change I could only launch ~ 6 participants with Connext, and after this change I could launch ~66 participants with Connext.

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9617/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5313/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7840/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9523/)

mjcarroll:
Does this have a noticeable impact on memory usage in `rmw_connext`?

clalancette:
> Does this have a noticeable impact on memory usage in `rmw_connext`?

Good question.

I ran the standard talker and listener demo with and without this change.  I also ran the large node test [here](https://github.com/clalancette/many_nodes/blob/master/launch/many_nodes-launch.py) (with the # of nodes restricted to 6, for an apples-to-apples comparison).  Here are the results:

| Binary | Label | Without increasing (kb) | With increasing (kb) | Difference (kb) |
| -------- | -------- | ------------------------ | --------------------- | ------------------- |
| talker | RSS | 42660 |43612 | +950 |
| talker | VIRT | 1425804 | 1426928 | +1124 |
| listener | RSS | 42924 | 44208 | +1284 |
| listener | VIRT | 1352028 |1353204 | +1176 |
| many_nodes | RSS | 129672 | 134564 | +4892 |
| many_nodes | VIRT | 7698464 | 7707076 | +8612 |

It's not totally consistent, but it looks like this change introduces ~800 - 900 kb more RSS memory per participant.

clalancette:
Another CI, since this is a bit stale now:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9766/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5386/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7935/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9631/)

clalancette:
All right, all of the failures are also on the nightlies.  So I'm going to merge this.  Thanks for the reviews.

PR #395, opened 2020-03-12
Fix memory leak in temporary list of node names

jacobperron:
Pointed out in https://github.com/ros2/rmw_connext/pull/362/files#r391680021

I believe `rcutils_string_array_fini` should handle freeing the memory.

### DIFF
--- a/rmw_connext_shared_cpp/src/node_names.cpp
+++ b/rmw_connext_shared_cpp/src/node_names.cpp
@@ -178,10 +178,7 @@ get_node_names(
 
   for (auto i = 0; i < named_nodes_num; ++i) {
     node_names->data[i] = rcutils_strdup(tmp_names_list.data[i], allocator);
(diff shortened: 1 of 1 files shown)


jacobperron:
> On a side note: The whole logic could be simplified a lot. e.g.: by first getting the count of available names and then in another loop populating the output array (without the need of a temporal output array).

Agreed. If I have a moment I can propose a simplified solution.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9629/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5323/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7853/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9533/)

PR #397, opened 2020-03-17
Add rmw_*_event_init() functions

mm318:
Related to https://github.com/ros2/ros2/issues/822

This adds the `rmw_publisher_event_init()` and `rmw_subscription_event_init()` functions to `rmw_connext`, moved from `rmw`, so that these functions will be able to check for whether or not the given type of `rmw_event_type_t` is supported by the chosen middleware.

### DIFF
--- a/rmw_connext_cpp/src/rmw_event.cpp
+++ b/rmw_connext_cpp/src/rmw_event.cpp
@@ -20,17 +20,6 @@
 
 extern "C"
 {
-
-/// Initialize a rmw publisher event.
-/**
- * \param[in|out] rmw_event to initialize
- * \param publisher to initialize with
- * \param event_type for the event to handle
- * \return `RMW_RET_OK` if successful, or
- * \return `RMW_RET_INVALID_ARGUMENT` if invalid argument, or
- * \return `RMW_RET_UNSUPPORTED` if event_type is not supported, or
- * \return `RMW_RET_ERROR` if an unexpected error occurs.
- */
 rmw_ret_t
 rmw_publisher_event_init(
   rmw_event_t * rmw_event,
@@ -45,16 +34,6 @@ rmw_publisher_event_init(
     event_type);
 }
 
-/// Initialize a rmw subscription event.
-/**
- * \param[in|out] rmw_event to initialize
- * \param subscription to initialize with
- * \param event_type for the event to handle
- * \return `RMW_RET_OK` if successful, or
- * \return `RMW_RET_INVALID_ARGUMENT` if invalid argument, or
- * \return `RMW_RET_UNSUPPORTED` if event_type is not supported, or
- * \return `RMW_RET_ERROR` if an unexpected error occurs.
- */
 rmw_ret_t
 rmw_subscription_event_init(
   rmw_event_t * rmw_event,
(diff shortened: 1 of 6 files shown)


rmw_connext_cpp/src/rmw_event.cpp
```diff
@@ -20,6 +20,55 @@
...
+/// Initialize a rmw publisher event.
+/**
+ * \param[in|out] rmw_event to initialize
+ * \param publisher to initialize with
+ * \param event_type for the event to handle
+ * \return `RMW_RET_OK` if successful, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if invalid argument, or
+ * \return `RMW_RET_UNSUPPORTED` if event_type is not supported, or
+ * \return `RMW_RET_ERROR` if an unexpected error occurs.
+ */
```
wjwwood: The docstrings here are redundant are they not?

rmw_connext_shared_cpp/src/event.cpp
```diff
@@ -20,6 +20,35 @@
...
+  RMW_CHECK_ARGUMENT_FOR_NULL(identifier, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_ARGUMENT_FOR_NULL(rmw_event, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_ARGUMENT_FOR_NULL(topic_endpoint_impl_identifier, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_ARGUMENT_FOR_NULL(data, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    topic endpoint,
+    topic_endpoint_impl_identifier,
+    identifier,
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  if (!is_event_supported(event_type)) {
```
wjwwood: Did `is_event_supported()` already exist? I don't see it in this pr.

mm318: Yep, it already exists, declared in `event_converter.hpp`.

rmw_connext_cpp/src/rmw_event.cpp
mm318: I agree. I saw it was done for `rmw_take_event()`, so I did it for `rmw_publisher_event_init()` and `rmw_subscription_event_init()` as well. I'll remove all of them.

rmw_connext_shared_cpp/include/rmw_connext_shared_cpp/event.hpp
```diff
@@ -22,6 +22,26 @@
...
+ * \param data to initialize with
+ * \param event_type for the event to handle
+ * \return `RMW_RET_OK` if successful, or
+ * \return `RMW_RET_INVALID_ARGUMENT` if invalid argument, or
+ * \return `RMW_RET_UNSUPPORTED` if event_type is not supported, or
+ * \return `RMW_RET_ERROR` if an unexpected error occurs.
+ */
+RMW_CONNEXT_SHARED_CPP_PUBLIC
+rmw_ret_t
+__rmw_init_event(
```
ivanpauno: A bit unrelated because it's extensively used in this package, function names starting with a leading underscore or containing `__` should be avoided, as those names are reserved by the standard.
See https://github.com/ros2/rcl/pull/597#discussion_r392422395.

mm318: We can do a separate pull request that changes all the function names in this repo consistently.

PR #401, opened 2020-04-02
[WIP] Add support for security logging plugin

kyrofa:
If a `logging.xml` file is included in the security files, enable the security logging plugin in Connext using the properties contained in said file. Refer to table 10.1 in the [Connext 5.3.1 docs](https://community.rti.com/static/documentation/connext-dds/5.3.1/doc/manuals/connext_dds/dds_security/RTI_SecurityPlugins_GettingStarted.pdf) for the properties being used here.


rmw_connext_shared_cpp/test/test_security_logging.cpp
```diff
@@ -0,0 +1,212 @@
...
+
+TEST_F(SecurityLoggingTest, test_all)
+{
+  std::string xml_file_path = write_logging_xml(
+    "<file>foo</file>\n"
+    "<verbosity>ERROR</verbosity>\n"
+    "<distribute>true</distribute>\n"
+    "<qos>\n"
+    "  <depth>10</depth>\n"
+    "</qos>");
```
kyrofa, line 203: @mikaelarguedas this is using the original format. I'll switch to the nested one as soon as I get some clarification in the design doc.

kyrofa: This is now done.

rmw_connext_shared_cpp/src/node.cpp
```diff
@@ -215,6 +217,11 @@ create_node(
       RMW_SET_ERROR_MSG("failed to allocate memory for 'perm_fn'");
       goto fail;
     }
+    logging_fn = rcutils_join_path(srp, "logging.xml", allocator);
```
kyrofa, line 220: @mikaelarguedas if this file wasn't located alongside the other security files, we would probably need to modify RCL to find it for us, which limits the benefit of doing any of this parsing at the RMW level (since it requires an RCL API update anyway).

mikaelarguedas: If this file is meant to be ROS specific and contain only ROS concepts. Maybe it makes more sense to parse it in RCL (where the ROS concepts live). This would allow to avoid any duplication of parsing code and each RMW implementation (DDS or not) can use the parsed information as it sees fit.

This stands regardless of where the file is located. However, having that logic in RCL would enable more easily support for the case the file is at another location

kyrofa:
Going to shoot for a simpler MVP.

PR #404, opened 2020-04-06
Add basic support for security logging plugin

kyrofa:
Support setting the security logging properties in Connext by exposing three environment variables:

- `ROS_SECURITY_LOG_FILE`: Log security events to the provided file path
- `ROS_SECURITY_LOG_PUBLISH`: Publish security events to DDS topic
- `ROS_SECURITY_LOG_VERBOSITY`: Control verbosity of logged events

cc @mikaelarguedas and @artivis


rmw_connext_shared_cpp/src/security_logging.cpp
```diff
@@ -0,0 +1,140 @@
...
+}
+}  // namespace
+
+rmw_ret_t apply_security_logging_configuration(DDS::PropertyQosPolicy & policy)
+{
+  const char * env_value;
+
+  // Handle logging to file
+  if (!get_env(log_file_variable_name, &env_value)) {
+    return RMW_RET_ERROR;
```
artivis, line 100: Does that mean that the env var *must* be defined? ([ret code goto fail](https://github.com/ros2/rmw_connext/pull/404/files#diff-2469e9c52ef0878455668e9233e96b0aR281-R282))
Same question for the other two en vars.

kyrofa: No, `rcutils_get_env()` will [set it to an empty string](https://github.com/ros2/rcutils/blob/master/include/rcutils/get_env.h#L32) if the environment variable is not set.

artivis: But what will return `get_env` if it not defined (true/false)?

kyrofa: `get_env()` only returns false if `rcutils_get_env()` returns non-NULL, which [only happens in the case of an error](https://github.com/ros2/rcutils/blob/master/include/rcutils/get_env.h#L34).

artivis: Ok that clarify the situation thanks.

rmw_connext_shared_cpp/src/security_logging.cpp
```diff
@@ -0,0 +1,140 @@
...
+    }
+  }
+
+  // Handle log verbosity
+  if (!get_env(log_verbosity_variable_name, &env_value)) {
+    return RMW_RET_ERROR;
+  }
+  if (strlen(env_value) > 0) {
+    const char * level;
+    if (!string_to_verbosity_level(env_value, &level)) {
```
artivis, line 126: Both `log_file` and `log_publish` values are not checked for correctness (e.g. is a bool) and this is, I assume, delayed to Connext. But here you are checking that the value is within the enum. How about returning instead a default error-level and let Connext do the checking as well for consistency?

kyrofa: Good point. I'd prefer to be consistent the other direction though and just enforce true/false here. I'm not sure what connext supports, but different vendors may support different things. We should limit that to what we know is supported.

artivis: > We should limit this to what we know is supported across them.

Which might be vastly different `true/True/on/On/1/One/tRuE/verdadera` and whatnot. Furthermore, I'd argue that this is a wrapper, let keep it thin and not duplicate any logic.

kyrofa: > Which might be vastly different true/True/on/On/1/One/tRuE/verdadera

Right, that's exactly my point. Ideally one would be able to set these environment variables (or define the XML that will follow) not caring what RMW is being used. If we don't limit what we support here the possible values will diverge depending on RMW.

artivis: gotcha :+1:

kyrofa: Alright I added this validation for booleans. The file path still seems reasonable to be handled by the DDS vendor, not least because dealing with file paths in a cross platform way here seems error-prone.

ruffsl:
Questions:
* Is there a particular reason for exposing these logging options using environment variables rather than through the vendor QoS XML settings file?
  * Is it to harmonize with the other ros security envs?
  * Is setting the vendor QoS XML too out of band?
* Are we only defining these environment variables in this rmw because connext is the only DDS vendor with a security logging plugin?
  * They seem like they should be declared in rcl.
* How does this effect the behavior of stdout/stderr with respect to security log events?
  * I suppose this will be vendor specific.

kyrofa:
> Is there a particular reason for exposing these logging options using environment variables rather than through the vendor QoS XML settings file?
> - Is it to harmonize with the other ros security envs?
> - Is setting the vendor QoS XML too out of band?

Yeah, you got it. We want the user to be able to enable/disable logging on the ROS end without having to care about the DDS implementation they're using.

> Are we only defining these environment variables in this rmw because connext is the only DDS vendor with a security logging plugin?
> - They seem like they should be declared in rcl.

No, we propose doing the same in rmw_fastrtps. Long-term this should indeed be in RCL, but we felt it needed more design before we got to that point. Security logging is actually something configured on a per-participant basis, so we have an XML file format designed for this that we were thinking would be generated by the SROS2 utilities, but the context work and introduction of enclaves has changed that a bit and we need to revisit. In the short-term, some environment variables to allow people to start playing with this functionality while we properly design the real deal seemed the easiest path forward.

Would you prefer to see this added to RCL and shoved into rmw_security_options_t, even knowing that this isn't the final story?

ruffsl:
> Would you prefer to see this added to RCL and shoved into rmw_security_options_t, even knowing that this isn't the final story?

It might be nice to introduce it into rcl, even its beta, as even then other rmw vendors could be added to the beta feature mid release. I'd like to rework `rmw_security_options_t` in the next release anyway to see if it could handle PEM files and passphrases as well, so that something like `rcl_security` could load from more exotic keystore sources.

@ivanpauno or @mikaelarguedas might have more of an option on pushing these log env options up into `rcl` instead.

kyrofa:
> I'd like to rework rmw_security_options_t in the next release anyway

Same here, but note that I'm trying to get this stuff into Foxy.

jacobperron:
If you think any new additions to `rmw_security_options_t` are likely to change significantly in G-turtle, then I don't think it's worth moving this to rcl. My reasoning is it avoid breaks for any users who might end up using the new API introduced in Foxy. The env variables can be reused (or easily deprecated) while moving implementation to rcl in G-turtle.

kyrofa:
Thank you @jacobperron, you outlined my position in far less words than I did :wink: . Okay, this is ready for a more thorough review, then. Once we're happy with it I can update https://github.com/ros2/rmw_fastrtps/pull/362 to match.

rmw_connext_shared_cpp/test/test_security_logging.cpp
```diff
@@ -0,0 +1,194 @@
...
+{
+  return lookup_property_value(policy, distribute_enable_property_name);
+}
+
+class SecurityLoggingTest : public ::testing::Test
+{
+public:
+  void SetUp()
+  {
+    custom_setenv(log_file_variable_name, "");
```
jacobperron, line 82: There's a difference between the environment variable being set to `""` and not being set at all.

I suggest `SetUp()` ensures that the variables are *unset*. And then we can have additional tests for when variables are set to empty strings.

jacobperron:
This PR needs a rebase to get the latest changes upstream so that CI will pass.

rmw_connext_shared_cpp/test/test_security_logging.cpp
kyrofa: Is there actually a difference? Unless I'm mistaken it wouldn't increase coverage at all here, but it _would_ require more platform-specific ifdefs.

> And then we can have additional tests for when variables are set to empty strings.

As far as I can tell, such adding more tests to test against empty strings or not set would in fact be tests against [`rcutils_get_env()`](https://github.com/ros2/rcutils/blob/master/include/rcutils/get_env.h#L32), which claims to return an empty string either way. Those types of tests should probably exist in rcutils, not here.

jacobperron: ah, I see. Since rcutils is returning an empty string either way, disregard this comment.

rmw_connext_shared_cpp/include/rmw_connext_shared_cpp/security_logging.hpp
```diff
@@ -0,0 +1,24 @@
...
+// limitations under the License.
+
+#ifndef RMW_CONNEXT_SHARED_CPP__SECURITY_LOGGING_HPP_
+#define RMW_CONNEXT_SHARED_CPP__SECURITY_LOGGING_HPP_
+
+#include <rmw/ret_types.h>
+
+#include "rmw_connext_shared_cpp/ndds_include.hpp"
+
+rmw_ret_t apply_security_logging_configuration(DDS::PropertyQosPolicy & policy);
```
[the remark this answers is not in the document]
kyrofa: GitHub doesn't support "reacting" with a face-palm emoji. So here: :facepalm: . On it.

rmw_connext_shared_cpp/src/security_logging.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include <rmw/error_handling.h>
+#include <rmw/qos_profiles.h>
+#include <rmw/types.h>
+
+#include "rmw_connext_shared_cpp/security_logging.hpp"
+
+namespace
+{
+// Environment variable names
+const char log_file_variable_name[] = "ROS_SECURITY_LOG_FILE";
```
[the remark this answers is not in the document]
kyrofa: Makes the API a bit larger, but you're right: the tradeoff is duplication. No strong feelings, I'll move them.

mikaelarguedas:
A bit late to the party but just to restate publically what happened in passed discussions.

End goal is to have a single entry point for configuring logging, that would be an XML file containing only ROS concepts (so not vendor specific and even not DDS-specific).
The code for parsing that xml and populating the relevant the datastructure should be implemented only once and reused by all implementations supporting secure logging.
Removing any requirement on environment variables in RCL is desired. The added ones here are temporary and do not impact the refactoring of environment variables in rcl into another package (ala https://github.com/ros2/rcl/issues/545)

These were too big of changes to be made last minute for Foxy, this PR and corresponding rmw_fastrtps PR are minimal self contained implementation that allow to configure some parts of the logging plugin to allow a minimal use. This was meant as an MVP not touching asny public API, this is why all new functions and defines are private and why the code is duplicated between this PR and the rmw_fastrtps one.


+1 for no public API change and not change to rmw_security_options this time around.
Adding a TODO(security-wg) above all the temporary hardcoded strings and constants would be good to remind any reader that this is temporary and meant to be (re)factored out.


@kyrofa I thought this change came with an sros2 change to add the relevant pure DDS logging topic to the permissions. Is this change on a branch somewhere to be tested along this ?
Is there a branch from test_security allowing to validate that this work ? or is this something we're aiming for later on ?

I haven't been able to test this but hope to do so and review the PRs by tomorrow

rmw_connext_shared_cpp/src/security_logging.cpp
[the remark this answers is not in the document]
jacobperron: Up to you, I don't think it needs to change. I'm not sure if they make sense as public API :man_shrugging:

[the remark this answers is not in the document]
kyrofa: Very good.

rmw_connext_shared_cpp/src/security_logging.cpp
```diff
@@ -0,0 +1,152 @@
...
+const char log_publish_variable_name[] = "ROS_SECURITY_LOG_PUBLISH";
+const char log_verbosity_variable_name[] = "ROS_SECURITY_LOG_VERBOSITY";
+
+// Connext property names. These are being taken from table 10.1 of
+// https://community.rti.com/static/documentation/connext-dds/5.3.1/doc/manuals/connext_dds/dds_security/RTI_SecurityPlugins_GettingStarted.pdf
+const char log_file_property_name[] = "com.rti.serv.secure.logging.log_file";
+const char distribute_enable_property_name[] = "com.rti.serv.secure.logging.distribute.enable";
+const char verbosity_property_name[] = "com.rti.serv.secure.logging.log_level";
```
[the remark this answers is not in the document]
kyrofa: Given that we're leaving the others, I'll leave these alone at least for now.

rmw_connext_shared_cpp/include/rmw_connext_shared_cpp/security_logging.hpp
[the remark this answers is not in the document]
kyrofa: Alright, this has been added.

kyrofa:
> This PR needs a rebase to get the latest changes upstream so that CI will pass.

Done.

kyrofa:
> Adding a TODO(security-wg) above all the temporary hardcoded strings and constants would be good to remind any reader that this is temporary and meant to be (re)factored out.

Good idea, just pushed that.

> @kyrofa I thought this change came with an sros2 change to add the relevant pure DDS logging topic to the permissions. Is this change on a branch somewhere to be tested along this?

I'm working on that now, so not yet. However, this doesn't depend on that so much as it depends on this.

> Is there a branch from test_security allowing to validate that this work ? or is this something we're aiming for later on?

No, but I fully intend to write one, particularly to ensure RMWs that support logging work the same way. I just need to learn my way around them. Other than the unit tests in this PR to ensure we're turning the right knobs, actually ensuring it turns into Connext logging as expected is currently a manual affair. The test_security addition should not be subject to the feature freeze like this is, however.

rmw_connext_shared_cpp/src/security_logging.cpp
```diff
@@ -0,0 +1,154 @@
...
+  const char * const level;
+} supported_verbosities[] =
+{
+  {"EMERGENCY", "0"},
+  {"ALERT", "1"},
+  {"CRITICAL", "2"},
+  {"ERROR", "3"},
+  {"WARNING", "4"},
+  {"NOTICE", "5"},
+  {"INFORMATIONAL", "6"},
```
mikaelarguedas, line 49: TBH this has been pretty unexpected and I had to go in the code to figure out what value to put to get this working.
I epected one of the following to work:
- what's in the spec:
  - INFO_LEVEL
  - INFORMATIONAL_LEVEL
- What I saw in most DDS examples (Connext, CoreDX): numbers between 0 and 7
- The ROS severity levels https://github.com/ros2/rcutils/blob/a357b67e7d9f00f8fbfe306491438cd1e8ee5219/rcutils/logging.py#L17

And NONE worked...

I believe we need those to be both more visible, with a more informative error message than:
`failed to set security logging verbosity: 7 is not a supported verbosity`
We need to either accept more values, or pick ones that are already agreed on elsewhere, either the numbers or string matching the

rmw_connext_shared_cpp/src/node.cpp
```diff
@@ -21,6 +21,7 @@
 #include "rmw_connext_shared_cpp/ndds_include.hpp"
 #include "rmw_connext_shared_cpp/node.hpp"
 #include "rmw_connext_shared_cpp/types.hpp"
+#include "rmw_connext_shared_cpp/security_logging.hpp"
```
mikaelarguedas, line 24: Nit: alphabetical order

rmw_connext_shared_cpp/src/security_logging.cpp
```diff
@@ -0,0 +1,154 @@
...
+  // Overwrite existing properties, so remove it if it already exists
+  DDS::PropertyQosPolicyHelper::remove_property(policy, property_name);
+
+  auto status = DDS::PropertyQosPolicyHelper::add_property(
+    policy,
+    property_name,
+    value,
+    DDS::BOOLEAN_FALSE);
+
+  return status == DDS::RETCODE_OK;
```
mikaelarguedas, line 83: Nit: operands on left side
```suggestion
  return DDS::RETCODE_OK == status;
```

rmw_connext_shared_cpp/src/security_logging.cpp
kyrofa: Both Fast RTPS and Connext follow the enum in section 9.6 of DDS-Security... mostly. Connext uses the numeric value of the enum in 5.3, but deprecated that in v6 to use the strings instead, sans the "_LEVEL". Fast RTPS uses the "_LEVEL." However, given that:

1. We want all RMWs to work the same way, even given that this is temporary, and
2. We want this to be more of a ROS abstraction than a DDS-level thing

We settled on the more clear and less "go read the spec" ish option. You're right though-- would it help to simply print all the supported verbosities in the error message?

kyrofa: Also, I'd be happy to use the ROS verbosities instead of that seems more clear. There aren't quite as many levels though so we'll need to map conservatively. Maybe:

| ROS verbosity | DDS verbosity |
|---|---|
| FATAL | EMERGENCY |
| (also ERROR) | ALERT |
| (also ERROR) | CRITICAL |
| ERROR | ERROR |
| WARN | WARNING |
| INFO | NOTICE |
| DEBUG | INFORMATIONAL |

mikaelarguedas: I don't feel very strongly either way as this is temporary, as long as we make it more visible.
For the longer term, we'll need to decide what to put in the XML file exposed to users.
As that XML file aim to have only ROS concepts, maybe we should use ROS verbosity levels in the XML.
So for consistency I'd say let's use that here too.
The mapping in the table above looks good to me :+1:

rmw_connext_shared_cpp/src/security_logging.cpp
```diff
@@ -0,0 +1,140 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <rcutils/get_env.h>
+#include <rmw/error_handling.h>
+#include <rmw/qos_profiles.h>
+#include <rmw/types.h>
+
+#include "rmw_connext_shared_cpp/security_logging.hpp"
```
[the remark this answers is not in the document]
kyrofa: I think that assumes I use quotes for all includes and not angle brackets. The linters seem happy locally anyway.

rmw_connext_shared_cpp/src/security_logging.cpp
artivis: I can't really make my mind on whether we should map or not. I see the benefit of sticking to ROS verbosities but can't yet foresee the cons of having less levels. And the specs does not help much here, from section 9.6:
```
    EMERGENCY_LEVEL,    // System is unusable. Should not continue use.
    ALERT_LEVEL,        // Should be corrected immediately
    CRITICAL_LEVEL,     // A failure in primary application.
    ERROR_LEVEL,        // General error conditions
    WARNING_LEVEL,      // May indicate future error if action not taken.
    NOTICE_LEVEL,       // Unusual, but nor erroneous event or condition.
    INFORMATIONAL_LEVEL, // Normal operational. Requires no action.
    DEBUG_LEVEL
```
I guess either way is fine for now so we should prefer the most user friendly solution and that'd be using ROS verbosities.

kyrofa: Alright, just pushed a change to use ROS verbosities, as well as a nicer error message when an unsupported once is selected.

rmw_connext_shared_cpp/CMakeLists.txt
```diff
@@ -51,6 +51,7 @@ add_library(
   src/node_info_and_types.cpp
   src/service_names_and_types.cpp
   src/topic_names_and_types.cpp
+  src/security_logging.cpp
```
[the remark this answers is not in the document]
kyrofa: Fixed.

rmw_connext_shared_cpp/src/node.cpp
kyrofa: Fixed.

kyrofa:
I was just able to confirm that logging to DDS works via the RTI admin console.

kyrofa:
(https://github.com/ros2/rmw_fastrtps/pull/362 has been updated to function the same way as this PR)

kyrofa:
CI for rmw_connext_cpp and rmw_connext_shared_cpp:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10359/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5876/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8436/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10257/)

mikaelarguedas:
> The Windows warnings appear to be coming from rcutils' get_env.h, not this PR

The fact that these warnings appear in this PR that uses `rcutils_get_env` and they are not present on master when multiple other packages use that same function is suspicious. It looks like this PR is introducing these warnigs and we should find a fix to not do that.

kyrofa:
I was hoping maybe the visibility macro was the hidden magic here, but apparently not:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10381/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8464/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10278/)

Note that https://github.com/ros2/rmw_fastrtps/pull/362, which is super similar code, does not have a warning.

sloretz:
> Note that ros2/rmw_fastrtps#362, which is super similar code, does not have a warning. I'm at a loss here. @sloretz any chance you have an idea?

No idea. Maybe the issue can be avoided for now by using the `rcpputils` version? https://github.com/ros2/rcpputils/blob/master/include/rcpputils/get_env.hpp

rmw_connext_shared_cpp/src/security_logging.cpp
```diff
@@ -0,0 +1,201 @@
...
+};
+
+bool severity_names_str(char buffer[], size_t buffer_size)
+{
+  size_t offset = 0;
+
+  size_t severity_count = sizeof(verbosity_mapping) / sizeof(verbosity_mapping[0]);
+  for (size_t i = 0; i < (severity_count - 1); ++i) {
+    RCUTILS_LOG_SEVERITY severity = verbosity_mapping[i].ros_severity;
+    offset += rcutils_snprintf(
```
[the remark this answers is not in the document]
kyrofa: Good catch, and I agree on the fix. Done.

rmw_connext_shared_cpp/src/security_logging.cpp
```diff
@@ -0,0 +1,201 @@
...
+      g_rcutils_log_severity_names[severity]);
+    if (offset >= buffer_size) {
+      return false;
+    }
+  }
+
+  offset += rcutils_snprintf(
+    buffer + offset, buffer_size - offset, "or %s",
+    g_rcutils_log_severity_names[verbosity_mapping[severity_count - 1].ros_severity]);
+  return offset <= buffer_size;
```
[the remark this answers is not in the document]
kyrofa: Fixed as well.

kyrofa:
Alright, I've changed to using rcpputils, let's see if that resolves things:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10405/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5914/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8486/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10296/)

kyrofa:
Excellent, all green now!

PR #407, opened 2020-04-09
security-context -> enclave

mikaelarguedas:
Signed-off-by: Mikael Arguedas <mikael.arguedas@gmail.com>

### DIFF
--- a/rmw_connext_shared_cpp/src/node.cpp
+++ b/rmw_connext_shared_cpp/src/node.cpp
@@ -77,7 +77,7 @@ create_node(
     nullptr,
     0,
     "name=%s;namespace=%s;securitycontext=%s;",
-    name, namespace_, context->options.security_context) + 1;
+    name, namespace_, context->options.enclave) + 1;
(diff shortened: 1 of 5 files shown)


mikaelarguedas:
CI at https://github.com/ros2/sros2/pull/198#issuecomment-612017072

rmw_connext_shared_cpp/src/node.cpp
```diff
@@ -76,8 +76,8 @@ create_node(
   size_t length = std::snprintf(
     nullptr,
     0,
-    "name=%s;namespace=%s;securitycontext=%s;",
-    name, namespace_, context->options.security_context) + 1;
+    "name=%s;namespace=%s;enclave=%s;",
```
ruffsl, line 79: Oh, so this with the missing underscore was the string key that was used to shuttle around the context name in the QoS. I thought the the context name was still being used for other stuff, so I was wondering about updating of the graph API and discovery stuff. If the key was named `securitycontext` from the get go, then I think its fine to anix it purely for enclaves.

PR #408, opened 2020-04-10
Add support for taking a sequence of messages

mjcarroll:
Implementation of ros2/rmw#212

Signed-off-by: Michael Carroll <michael@openrobotics.org>

### DIFF
--- a/rmw_connext_cpp/src/rmw_take.cpp
+++ b/rmw_connext_cpp/src/rmw_take.cpp
@@ -152,7 +152,7 @@ _take(
   RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
     subscription handle,
     subscription->implementation_identifier, rti_connext_identifier,
-    return RMW_RET_ERROR)
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION)
 
   if (!ros_message) {
     RMW_SET_ERROR_MSG("ros message handle is null");
@@ -214,16 +214,26 @@ _take_sequence(
   (void) allocation;
   if (!subscription) {
     RMW_SET_ERROR_MSG("subscription handle is null");
-    return RMW_RET_ERROR;
+    return RMW_RET_INVALID_ARGUMENT;
   }
   RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
     subscription handle,
     subscription->implementation_identifier, rti_connext_identifier,
-    return RMW_RET_ERROR)
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION)
 
   if (!message_sequence) {
     RMW_SET_ERROR_MSG("message sequence handle is null");
-    return RMW_RET_ERROR;
+    return RMW_RET_INVALID_ARGUMENT;
+  }
+
+  if (!message_info_sequence) {
+    RMW_SET_ERROR_MSG("message info sequence handle is null");
+    return RMW_RET_INVALID_ARGUMENT;
+  }
+
+  if (!taken) {
+    RMW_SET_ERROR_MSG("taken handle is null");
+    return RMW_RET_INVALID_ARGUMENT;
   }
 
   if (count > message_sequence->capacity) {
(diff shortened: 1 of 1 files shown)


rmw_connext_cpp/src/rmw_take.cpp
```diff
@@ -195,6 +202,137 @@ _take(
...
+{
+  (void) allocation;
+  if (!subscription) {
+    RMW_SET_ERROR_MSG("subscription handle is null");
+    return RMW_RET_ERROR;
+  }
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    subscription handle,
+    subscription->implementation_identifier, rti_connext_identifier,
+    return RMW_RET_ERROR)
```
wjwwood, line 222: This should be `RMW_RET_INCORRECT_RMW_IMPLEMENTATION` right?

https://github.com/ros2/rmw/blob/974ef929e22569ed811edbbf70ba85718dbcbc02/rmw/include/rmw/ret_types.h#L32-L37

rmw_connext_cpp/src/rmw_take.cpp
```diff
@@ -195,6 +202,137 @@ _take(
...
+  const rmw_subscription_t * subscription,
+  size_t count,
+  rmw_message_sequence_t * message_sequence,
+  rmw_message_info_sequence_t * message_info_sequence,
+  size_t * taken,
+  rmw_subscription_allocation_t * allocation)
+{
+  (void) allocation;
+  if (!subscription) {
+    RMW_SET_ERROR_MSG("subscription handle is null");
```
wjwwood, line 216: This should be `RMW_RET_INVALID_ARGUMENT` right?

https://github.com/ros2/rmw/blob/974ef929e22569ed811edbbf70ba85718dbcbc02/rmw/include/rmw/ret_types.h#L32-L37

rmw_connext_cpp/src/rmw_take.cpp
```diff
@@ -195,6 +202,137 @@ _take(
...
+    if (!ignore_sample) {
+      rcutils_uint8_array_t cdr_stream = rcutils_get_zero_initialized_uint8_array();
+
+      cdr_stream.buffer_length = dds_messages[ii].serialized_data.length();
+      cdr_stream.buffer_capacity = dds_messages[ii].serialized_data.length();
+      cdr_stream.buffer = reinterpret_cast<uint8_t *>(&dds_messages[ii].serialized_data[0]);
+
+      if (callbacks->to_message(
+          &cdr_stream,
+          message_sequence->data[*taken]))
```
wjwwood, line 317: nitpick: I'd do it like this:

```c++
      if (callbacks->to_message(&cdr_stream, message_sequence->data[*taken]))
      {
        rmw_gid_t * sender_gid = &message_info_sequence->data[*taken].publisher_gid;
        sender_gid->implementation_identifier = rti_connext_identifier;
        memset(sender_gid->data, 0, RMW_GID_STORAGE_SIZE);
        auto detail = reinterpret_cast<ConnextPublisherGID *>(sender_gid->data);
        detail->publication_handle = sample_info.publication_handle;
        (*taken)++;
      }
```

I don't think it's long enough to need wrapping.

mjcarroll: You are correct, fixed.

rmw_connext_cpp/src/rmw_take.cpp
```diff
@@ -83,21 +103,8 @@ take(
   if (!sample_info.valid_data) {
     // skip sample without data
     ignore_sample = true;
-  } else if (ignore_local_publications) {
-    // compare the lower 12 octets of the guids from the sender and this receiver
-    // if they are equal the sample has been sent from this process and should be ignored
```
jacobperron, line 88: It would be nice to retain this comment with the moved logic.

mjcarroll: Done in 1161405

PR #411, opened 2020-04-17
fix CMake warning about using uninitialized variables

dirk-thomas:
When invoking CMake with `--warn-uninitialized`.

### DIFF
--- a/rmw_connext_cpp/CMakeLists.txt
+++ b/rmw_connext_cpp/CMakeLists.txt
@@ -3,11 +3,11 @@ cmake_minimum_required(VERSION 3.5)
 project(rmw_connext_cpp)
 
 if(DEFINED ENV{CONNEXT_STATIC_DISABLE})
-  set(disable $ENV{CONNEXT_STATIC_DISABLE})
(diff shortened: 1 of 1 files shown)


rmw_connext_cpp/CMakeLists.txt
```diff
@@ -2,7 +2,12 @@ cmake_minimum_required(VERSION 3.5)
 
 project(rmw_connext_cpp)
 
-set(CONNEXT_STATIC_DISABLE $ENV{CONNEXT_STATIC_DISABLE}
+if(DEFINED ENV{CONNEXT_STATIC_DISABLE})
+  set(disable $ENV{CONNEXT_STATIC_DISABLE})
```
hidmic, line 6: @dirk-thomas nit: can we replace generic `disable` variable name with `_connext_static_disable` or `disable_static_connext` ?

dirk-thomas: Done: f8690e8f88f39c8a0d63f920f8fd0ee3d26fce52.

PR #414, opened 2020-04-24
More verbose error output when apply_patch.py fails

ivanpauno:
See https://github.com/ros2/build_farmer/issues/273 for more context.

From https://ci.ros2.org/job/ci_linux/10475/console#console-section-26, it seems that `connext_static_serialized_dataPlugin.h.patch` is being applied to `connext_static_serialized_data.cxx` ... which would be wrong.

The line that would confirm that isn't being shown in the logs, and I think it's because the message is being printed in `stdout`.

### DIFF
--- a/rmw_connext_cpp/bin/apply-patch.py
+++ b/rmw_connext_cpp/bin/apply-patch.py
@@ -18,6 +18,7 @@
 
 import argparse
 import re
+import sys
 
 # regular expression / pattern for patch header
 _hdr_pat = re.compile(r'^@@ -(\d+),?(\d+)? \+(\d+),?(\d+)? @@.*$')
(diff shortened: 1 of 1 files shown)


ivanpauno:
https://github.com/ros2/rmw_connext/pull/414/commits/60674d19efae7dbcd15c36be5f4ac11a93c88281 also prints the content of the files.

It might be a bit noisy, but it can also be helpful to figure out the problem.

rmw_connext_cpp/bin/apply-patch.py
```diff
@@ -79,7 +80,13 @@ def apply_patch(s, patch):
     try:
         content_out = apply_patch(content_in, content_patch)
     except Exception:
-        print(i, p, o)
+        print(i, p, o, file=sys.stderr)
```
dirk-thomas, line 83: How about making this a "full message" like: "Failed to generate '{o}' using input '{i}' and patch '{p}"?

ivanpauno: done

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10499/)

PR #415, opened 2020-04-27
Retry calling the idl_pp up to 10 times.

clalancette:
This is because of possible races in rtiddsgen_server.

In theory this will mitigate the problems investigated in https://github.com/ros2/build_farmer/issues/273 .  However, since we don't have a reliable reproducer, we are just going to have to try it out and see if it helps.

One additional thing that we may want to do is forcefully kill off `rtiddsgen_server` before trying again.  However, we:

1.  Don't have proof that that is necessary, and
1.  I don't know a good cross-platform way to do that

So I think this is a good start for now.

### DIFF
--- a/rmw_connext_cpp/bin/call-idl-pp.py
+++ b/rmw_connext_cpp/bin/call-idl-pp.py
@@ -1,0 +1,47 @@
+# This is a utility function to help generate consistent results from Connext rtiddsgen{_server}.
+# We've seen cases where it can generate header files that do not contain the proper 'extern "C"'
+# markings, which we believe is a race inside the generator itself.
+# To combat this, we try up to 10 times to generate, and ensure that the header files have the
+# correct markings.  Once they do, we consider it a success.
+
+import argparse
+import os
+import shutil
+import subprocess
+import sys
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--idl-pp', required=True)
+    parser.add_argument('--idl-file', required=True)
+    parser.add_argument('-d', required=True)
+    parser.add_argument('--max-tries', type=int, default=10)
+    args = parser.parse_args()
+
+    done = False
+    count = 0
+    while not done and count < args.max_tries:
+        # Remove and recreate the output target directory.  This ensures that previous failed
+        # attempts won't cause the next attempt to fail.
+        try:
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10526/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5984/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8572/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10391/)
* Windows Debug [Build Status](http://ci.ros2.org/job/ci_windows/10392/)

dirk-thomas:
* The script should print some indication that a retry is happening.
* Run linters on the file (and address their warnings).

rmw_connext_cpp/bin/call-idl-pp.py
```diff
@@ -0,0 +1,69 @@
...
+            pass
+        os.mkdir(args.d)
+
+        ret = subprocess.run(args=[args.idl_pp, '-language', 'C++', '-unboundedSupport',
+                                   args.idl_file, '-d', args.d])
+        if ret.returncode == 0:
+            with open(os.path.join(args.d, args.idl_file[:-4] + 'Plugin.h'), 'r') as infp:
+                for line in infp:
+                    if line.startswith('extern "C" {'):
+                        return ret.returncode
```
dirk-thomas, line 53: Maybe return a literal `0` here?

clalancette:
The warnings on Windows are unrelated, and fixed by https://github.com/ros2/rmw_connext/pull/416 .  CI is otherwise clean, so merging this.  In the long run, if this doesn't fix the problem, we'll revisit it.

dirk-thomas:
Example build failing all 10 retries: http://build.ros2.org/view/Fci/job/Fci__nightly-extra-rmw-release_ubuntu_focal_amd64/95/console It seems if it fails it fails deterministically in the build.

Maybe an alternative approach could be to fallback to the non-server generator if the server one is used and fails?

clalancette:
> Maybe an alternative approach could be to fallback to the non-server generator if the server one is used and fails?

Yeah, that's a good idea.  I'm going to come up with a patch that does that and adds even more debugging information as well.

PR #416, opened 2020-04-27
Cast size_t to DDS_Long explicitly.

hidmic:
This pull request aims to fix [a warning seen on nightly Windows builds](https://ci.ros2.org/view/nightly/job/nightly_win_rel/1533/msbuild/category.63474818/). Said warning was introduced by https://github.com/ros2/rmw_connext/pull/408. This was first observed when running CI for [ros2/launch#408](https://github.com/ros2/launch/pull/408#issuecomment-620162001). 

CI up to `rmw_connext` and `rmw_cyclonedds_cpp`:
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/10425/)

### DIFF
--- a/rmw_connext_cpp/src/rmw_take.cpp
+++ b/rmw_connext_cpp/src/rmw_take.cpp
@@ -283,7 +283,7 @@ _take_sequence(
   DDS::ReturnCode_t status = data_reader->take(
     dds_messages,
     sample_infos,
-    count,
+    static_cast<DDS_Long>(count),
     DDS::ANY_SAMPLE_STATE,
     DDS::ANY_VIEW_STATE,
     DDS::ANY_INSTANCE_STATE);
--- a/rmw_connext_cpp/src/rmw_take.cpp
+++ b/rmw_connext_cpp/src/rmw_take.cpp
@@ -248,6 +248,13 @@ _take_sequence(
     return RMW_RET_ERROR;
   }
 
+  if (count > std::numeric_limits<DDS_Long>::max()) {
+    RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
+      "cannot take %ld samples at once, limit is %d",
+      count, std::numeric_limits<DDS_Long>::max());
+    return RMW_RET_ERROR;
+  }
+
   ConnextStaticSubscriberInfo * subscriber_info =
     static_cast<ConnextStaticSubscriberInfo *>(subscription->data);
   if (!subscriber_info) {
--- a/rmw_connext_cpp/src/rmw_take.cpp
+++ b/rmw_connext_cpp/src/rmw_take.cpp
@@ -248,10 +248,10 @@ _take_sequence(
     return RMW_RET_ERROR;
   }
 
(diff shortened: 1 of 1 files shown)


rmw_connext_cpp/src/rmw_take.cpp
```diff
@@ -283,7 +283,7 @@ _take_sequence(
   DDS::ReturnCode_t status = data_reader->take(
     dds_messages,
     sample_infos,
-    count,
+    static_cast<DDS_Long>(count),
```
clalancette, line 286: So, going through RTI headers a bit, it looks to me like:

```
typedef RTICdrLong DDS_Long;
```

and

```
typedef RTI_INT32 RTICdrLong;
```

and

```
typedef signed int  RTI_INT32;
```

On all of our target platforms, a signed integer is 32-bits.  This means that on 64-bit platforms theoretically `count` (which is a size_t) could hold a number much larger than what will fit into the signed 32-bit integer.  Thus, if we are going to do this, I think we need to have a check much earlier on that count <= DDS_Long::max.  Thoughts?

mjcarroll: If all of the supported RMW implementations take something 32 bits (either signed or unsigned), does it even make sense to use a `size_t` here? 

 I suppose I mentally defaulted to that to represent the size of something.

clalancette: Yeah, I'm not sure.  I haven't done a comparison with the other RMW's to see what they take.

Given that we are past API/feature freeze for Foxy, though, I'd say leave the API as-is and add in a check at the beginning of the method.

hidmic: Good point, done in https://github.com/ros2/rmw_cyclonedds/pull/171/commits/fe8ed460edff790cf2551177e098230bb4aadf12.

hidmic:
Argh, MIN/MAX macro expansion... PTAL at https://github.com/ros2/rmw_connext/pull/416/commits/5527948e6f23bc9cdf10ff3bba80d9b813f72942.

hidmic:
Alright, going in!

rmw_connext_cpp/src/rmw_take.cpp
```diff
@@ -248,6 +248,13 @@ _take_sequence(
     return RMW_RET_ERROR;
   }
 
+  if (count > (std::numeric_limits<DDS_Long>::max)()) {
```
sloretz, line 251: Looks like this introduces a new compiler warning on linux https://ci.ros2.org/job/ci_linux/10551/gcc/new/source.bde60058-8945-4cc4-ae0d-abd08de67f16/#252

`comparison of integer expressions of different signedness: ‘size_t’ {aka ‘long unsigned int’} and ‘int’ [-Wsign-compare] 252 | if (count > (std::numeric_limits::max)()) { | `


Though it's odd that the warning says `(std::numeric_limits::max)()` without the template type

sloretz: https://github.com/ros2/rmw_connext/pull/420

PR #419, opened 2020-04-28
Correct rmw_take() return value.

hidmic:
A tiny fix. I believe this was introduced in #408, and compilers implicitly casted a `bool` literal into an `rmw_ret_t` value.


hidmic:
CI up to `rmw_connext_cpp`
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10571/)

hidmic:
Going in.

PR #421, opened 2020-04-30
Switch to always using rtiddsgen (not the server).

clalancette:
This is to avoid potential problems where we could connect
to an earlier iteration of the server, and generate things
we do not expect.  Since we know we are only ever generating
one IDL, we just use the normal rtiddsgen and then use the
server for generating "everything else".

### DIFF
--- a/rmw_connext_cpp/CMakeLists.txt
+++ b/rmw_connext_cpp/CMakeLists.txt
@@ -87,9 +87,10 @@ set(generated_files
 )
 add_custom_command(
   OUTPUT ${generated_files}
-  # We *explicitly* do not use Connext_DDSGEN_SERVER here.  That's so that we don't
-  # accidentally connect to another version of the server that might still be running
-  # from previous runs in rosidl_typesupport_connext_cpp.
+  # We *explicitly* do not use Connext_DDSGEN_SERVER here.  The other time we invoke the
+  # server is in rosidl_typesupport_connext_cpp, and that uses a different set of arguments.
+  # We don't want to accidentally connect to that other version of the server that might
+  # still be running.
   COMMAND ${PYTHON_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/bin/call-idl-pp.py" --idl-pp "${Connext_DDSGEN}" --idl-file "connext_static_serialized_data.idl" -d ${generated_directory}
   WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/resources"
   COMMENT "Generating serialized type support for RTI Connext (using '${Connext_DDSGEN}')"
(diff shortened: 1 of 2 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/10577/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6026/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/8613/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/10474/)

clalancette:
> Maybe, the same is happening when we generate typesupport from messages, but we aren't realizing because we don't patch the output of those generation steps.

Yeah, a further step here would be to kill off the server at the end of the job, which helps any "lingering" gen servers.  It also doesn't help in the case where two are running on the same machine.  But anyway, I *think* this patch will help for this failure.

rmw_connext_cpp/CMakeLists.txt
```diff
@@ -93,9 +87,12 @@ set(generated_files
 )
 add_custom_command(
   OUTPUT ${generated_files}
-  COMMAND ${PYTHON_EXECUTABLE} "${CMAKE_CURRENT_SOURCE_DIR}/bin/call-idl-pp.py" --idl-pp ${_idl_pp} --idl-file "connext_static_serialized_data.idl" -d ${generated_directory}
+  # We *explicitly* do not use Connext_DDSGEN_SERVER here.  That's so that we don't
+  # accidentally connect to another version of the server that might still be running
+  # from previous runs in rosidl_typesupport_connext_cpp.
```
dirk-thomas, line 92: Please add to this comment that the other instance is invoked with different arguments and they seem to leak into an invocation from here.

rmw_connext_cpp/bin/call-idl-pp.py
```diff
@@ -44,8 +44,10 @@ def main():
             pass
         os.mkdir(args.d)
 
-        ret = subprocess.run(args=[args.idl_pp, '-language', 'C++', '-unboundedSupport',
-                                   args.idl_file, '-d', args.d])
+        cmdline = [args.idl_pp, '-language', 'C++', '-unboundedSupport', args.idl_file, '-d',
```
dirk-thomas, line 47: Nitpick: wrap after `[`.

rmw_connext_cpp/CMakeLists.txt
clalancette: Done in ec5638a

rmw_connext_cpp/bin/call-idl-pp.py
clalancette: Changed in ec5638a.

clalancette:
:crossed_fingers: that this fixes the issue for real now.

PR #424, opened 2020-05-28
Add message lost event

ivanpauno:
See https://github.com/ros2/rmw/issues/226.

As described there, the new message lost event is triggered when either sample lost or sample rejected DDS status callbacks are triggered.
That aims to make the API less DDS dependent, and the difference between lost and rejected doesn't sound like important (if we care about that, we could add a `reason` member to the status).

[First commit](https://github.com/ros2/rmw_connext/commit/8f026b9b3ee4c98eb541a87ed081e355ec708fd0) does the changes to map one rmw event to more than one DDS statuses.
[Second commit](https://github.com/ros2/rmw_connext/commit/04c67408c279e0774abff9d4664d7c9a2889123c) actually adds the message lost event.


ivanpauno:
Rebased to squash and clean-up commit messages, without introducing code changes.
The PR is going to be rebase-merged with the current two commits if CI is green.

PR #427, opened 2020-06-19
Handle RMW_DEFAULT_DOMAIN_ID.

hidmic:
Precisely what the title says. Connected to https://github.com/ros2/rcl/pull/689.

### DIFF
--- a/rmw_connext_shared_cpp/src/node.cpp
+++ b/rmw_connext_shared_cpp/src/node.cpp
@@ -274,8 +274,12 @@ create_node(
     }
   }
 
+  // No custom handling of RMW_DEFAULT_DOMAIN_ID. Simply use a reasonable domain id.
   participant = dpf_->create_participant(
-    static_cast<DDS::DomainId_t>(domain_id), participant_qos, NULL,
+    static_cast<DDS::DomainId_t>(
+      domain_id != RMW_DEFAULT_DOMAIN_ID ? domain_id : 0u),
+    participant_qos,
+    NULL,
     DDS::STATUS_MASK_NONE);
   if (!participant) {
     RMW_SET_ERROR_MSG("failed to create participant");


ivanpauno:
why is this needed?
there's similar logic in rcl to avoid passing the default domain id IIRC

ivanpauno:
See https://github.com/ros2/rcl/pull/689/files#r443041939

rmw_connext_shared_cpp/src/node.cpp
```diff
@@ -274,8 +274,12 @@ create_node(
     }
   }
 
+  // No custom handling of RMW_DEFAULT_DOMAIN_ID. Simply use a reasonable domain id.
   participant = dpf_->create_participant(
-    static_cast<DDS::DomainId_t>(domain_id), participant_qos, NULL,
+    static_cast<DDS::DomainId_t>(
+      domain_id != RMW_DEFAULT_DOMAIN_ID ? domain_id : 0u),
```
wjwwood, line 280: nitpick: line wrap looks unnecessary.

hidmic:
> why is this needed? there's similar logic in rcl to avoid passing the default domain id IIRC

No there isn't. Or there shouldn't be. The bug that https://github.com/ros2/rcl/pull/689 fixed was forcing a `0` when `ROS_DOMAIN_ID=`. The point of having an invalid domain ID as default is for `rmw` implementations to decide what to do with it. We can discuss whether we want to have such sentinel or not.

hidmic:
CI up to `rcl` and `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11229/) (canary build)

ivanpauno:
> > why is this needed? there's similar logic in rcl to avoid passing the default domain id IIRC
> 
> No there isn't. Or there shouldn't be. The bug that [ros2/rcl#689](https://github.com/ros2/rcl/pull/689) fixed was forcing a `0` when `ROS_DOMAIN_ID=`. The point of having an invalid domain ID as default is for `rmw` implementations to decide what to do with it. We can discuss whether we want to have such sentinel or not.

It has been historically used to determine in `rcl` if `ROS_DOMAIN_ID` variable has to be check or not (see https://github.com/ros2/rcl/blob/72965a0e3fa9ad0af6d0fe301a1d02edc746a1fb/rcl/src/rcl/init.c#L146-L147).
 It has never been used for letting the rmw implementation decide what to do with the default domain id (same in foxy, eloquent, dashing, etc).

Not addressing https://github.com/ros2/rcl/pull/689#discussion_r443041939 will break rmw implementations not maintained by us.

hidmic:
> It has historically been used to determine in rcl if ROS_DOMAIN_ID variable has to be check or not

True. In `rcl`. This patch is making sure that `rmw_init()` with default options as provided by implementation does not fail.

ivanpauno:
> If an implementation does not handle RMW_DEFAULT_DOMAIN_ID, it is broken already.

Currently, rcl never passes RMW_DEFAULT_DOMAIN_ID, so it's not already broken.

hidmic:
> Currently, `rcl` never passes RMW_DEFAULT_DOMAIN_ID, so it's not already broken.

It is if the `rmw` can't handle it. And that's regardless of whether it's most common downstream package guards it from that situation or not.

PR #428, opened 2020-06-22
Finalize context if and only if it's shutdown

hidmic:
This pull request, albeit controversial, ensures `rmw_context_fini()` fails on a not shutdown context, as its API contract establishes (see https://github.com/ros2/rmw/pull/242).

### DIFF
--- a/rmw_connext_cpp/src/rmw_init.cpp
+++ b/rmw_connext_cpp/src/rmw_init.cpp
@@ -21,6 +21,11 @@
 
 #include "rmw_connext_cpp/identifier.hpp"
 
+struct rmw_context_impl_t {
+  // Shutdown flag
+  bool is_shutdown{false};
+};
+
 extern "C"
 {
 rmw_ret_t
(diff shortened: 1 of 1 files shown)


rmw_connext_cpp/src/rmw_init.cpp
```diff
@@ -105,12 +110,27 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
...
+  context->impl->is_shutdown = false;
   rmw_ret_t ret = rmw_init_options_copy(options, &context->options);
   if (RMW_RET_OK != ret) {
+    delete context->impl;
     return ret;
   }
-  return init();
+  ret = init();
+  if (RMW_RET_OK != ret) {
+    if (RMW_RET_OK != rmw_init_options_fini(&context->options)) {
```
ivanpauno, line 126: this seem unrelated to the PR.
IMO, it should be done in another PR, or in a separate commit.

hidmic: That is fair, yeah. I came across leaks and I couldn't help it.

hidmic:
CI up to `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11240/) (canary build)

ivanpauno:
> CI up to test_rmw_implementation against all RMW implementations:

I would run test including up to an upper layer, like test_rclcp, just in case.
At least in one platform.

PR #429, opened 2020-06-22
Avoid leaking init options in context.

hidmic:
Spin off form #428.

### DIFF
--- a/rmw_connext_cpp/src/rmw_init.cpp
+++ b/rmw_connext_cpp/src/rmw_init.cpp
@@ -116,7 +116,7 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
     goto fail;
   }
   return ret;
- fail:
+fail:
   if (RMW_RET_OK != rmw_init_options_fini(&context->options)) {
     RMW_SAFE_FWRITE_TO_STDERR(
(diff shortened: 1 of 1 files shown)


rmw_connext_cpp/src/rmw_init.cpp
```diff
@@ -106,11 +106,23 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
   context->instance_id = options->instance_id;
   context->implementation_identifier = rti_connext_identifier;
   context->impl = nullptr;
+
   rmw_ret_t ret = rmw_init_options_copy(options, &context->options);
   if (RMW_RET_OK != ret) {
-    return ret;
+    goto fail;
```
ivanpauno, line 112: I think that the point that was raised in `rmw_cyclonedds` PR is valid here too.
If `rmw_init_options_copy` failed, there shouldn't be a need to call `rmw_init_options_fini`.
If that's needed, it's because `rmw_init_options_copy` is not handling errors properly.

hidmic: See https://github.com/ros2/rmw_cyclonedds/pull/198#discussion_r444237586.

hidmic:
This has been superseded by #432.

PR #432, opened 2020-06-24
Ensure compliant init/shutdown API implementation.

hidmic:
To comply with https://github.com/ros2/rmw/pull/243.

### DIFF
--- a/rmw_connext_cpp/src/rmw_init.cpp
+++ b/rmw_connext_cpp/src/rmw_init.cpp
@@ -12,14 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#include "rmw/init.h"
-
+#include <cassert>
+#include <cstring>
 #include <memory>
 
 #include "rcutils/strdup.h"
 
 #include "rmw/error_handling.h"
 #include "rmw/impl/cpp/macros.hpp"
+#include "rmw/init.h"
+
 #include "rmw_connext_shared_cpp/init.hpp"
 
 #include "rcpputils/scope_exit.hpp"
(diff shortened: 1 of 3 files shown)


hidmic:
@ros-pull-request-builder retest this please

rmw_connext_cpp/src/rmw_init.cpp
```diff
@@ -111,24 +116,40 @@ rmw_init_options_fini(rmw_init_options_t * init_options)
...
+  if (NULL != context->implementation_identifier) {
+    RMW_SET_ERROR_MSG("expected a zero-initialized context");
+    return RMW_RET_INVALID_ARGUMENT;
+  }
+
+  auto cleanup = rcpputils::make_scope_exit(
+    [context]() {
+      delete context->impl;
+      *context = rmw_get_zero_initialized_context();
+    });
```
ivanpauno, line 139: can you move this after the line where the context impl is created?
or add a check before this, failing if `context->impl` wasn't a `nullptr`.

hidmic: So, `delete nullptr` is a no-op and this object is created at a point where we've assumed `context` is zero-initialized. As I said elsewhere, we can split this into two `scope_exit` instances, but your concern about `context->impl` not being `nullptr` is orthogonal. I'm not entirely convinced we have to deal with bogus code that breaks API preconditions. We can `assert` though.

ivanpauno: > We can assert though.

assert sounds fine.

>  I'm not entirely convinced we have to deal with bogus code that breaks API preconditions

If the extra check needed to avoid UB is only one line, I prefer adding it.

hidmic: Fair enough, see https://github.com/ros2/rmw_connext/pull/432/commits/c1eb006f475586bd64392eb5f85a469d40648762

hidmic:
CI up to `test_rmw_implementation` against all RMW implementations:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11320/)

hidmic:
Merging!

PR #439, opened 2020-07-07
Ensure compliant node construction/destruction API.

hidmic:
Connected to https://github.com/ros2/rmw/pull/249.

### DIFF
--- a/rmw_connext_shared_cpp/src/node.cpp
+++ b/rmw_connext_shared_cpp/src/node.cpp
@@ -476,12 +476,10 @@ destroy_node(const char * implementation_identifier, rmw_node_t * node)
     implementation_identifier,
     return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
 
-  auto node_info = static_cast<ConnextNodeInfo *>(node->data);
-  assert(node_info != nullptr);
-  auto participant = static_cast<DDS::DomainParticipant *>(node_info->participant);
-  assert(participant != nullptr);
-
   rmw_ret_t ret = RMW_RET_OK;
+  auto node_info = static_cast<ConnextNodeInfo *>(node->data);
+  auto participant = static_cast<DDS::DomainParticipant *>(node_info->participant);
+
   // This unregisters types and destroys topics which were shared between
   // publishers and subscribers and could not be cleaned up in the delete functions.
   if (participant->delete_contained_entities() == DDS::RETCODE_OK) {
--- a/rmw_connext_shared_cpp/src/node.cpp
+++ b/rmw_connext_shared_cpp/src/node.cpp
@@ -54,7 +54,7 @@ create_node(
     return NULL;
   }
 
-  int validation_result = -1;
+  int validation_result = RMW_NODE_NAME_VALID;
   rmw_ret_t ret = rmw_validate_node_name(name, &validation_result, nullptr);
   if (RMW_RET_OK != ret || RMW_NODE_NAME_VALID != validation_result) {
     const char * reason = RMW_RET_OK == ret ?
@@ -63,7 +63,7 @@ create_node(
     RMW_SET_ERROR_MSG_WITH_FORMAT_STRING("invalid node name: %s", reason);
     return NULL;
   }
-  validation_result = -1;
+  validation_result = RMW_NAMESPACE_VALID;
   ret = rmw_validate_namespace(namespace_, &validation_result, nullptr);
   if (RMW_RET_OK != ret || RMW_NAMESPACE_VALID != validation_result) {
     const char * reason = RMW_RET_OK == ret ?
--- a/rmw_connext_shared_cpp/src/node.cpp
+++ b/rmw_connext_shared_cpp/src/node.cpp
@@ -18,6 +18,7 @@
 #include "rcutils/filesystem.h"
 
 #include "rmw_connext_shared_cpp/guard_condition.hpp"
+#include "rmw_connext_shared_cpp/init.hpp"
 #include "rmw_connext_shared_cpp/ndds_include.hpp"
 #include "rmw_connext_shared_cpp/node.hpp"
 #include "rmw_connext_shared_cpp/security_logging.hpp"
@@ -26,6 +27,8 @@
 #include "rmw/allocators.h"
 #include "rmw/error_handling.h"
 #include "rmw/impl/cpp/macros.hpp"
+#include "rmw/validate_namespace.h"
+#include "rmw/validate_node_name.h"
 
 rmw_node_t *
 create_node(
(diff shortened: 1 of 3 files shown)


rmw_connext_shared_cpp/src/node.cpp
```diff
@@ -451,14 +486,10 @@ destroy_node(const char * implementation_identifier, rmw_node_t * node)
...
-  if (!node_info) {
-    RMW_SET_ERROR_MSG("node info handle is null");
-    return RMW_RET_ERROR;
-  }
+  assert(node_info != nullptr);
   auto participant = static_cast<DDS::DomainParticipant *>(node_info->participant);
-  if (!participant) {
-    RMW_SET_ERROR_MSG("participant handle is null");
-  }
+  assert(participant != nullptr);
```
ivanpauno, line 491: why assert?

rmw_connext_shared_cpp/src/node.cpp
```diff
@@ -451,14 +486,10 @@ destroy_node(const char * implementation_identifier, rmw_node_t * node)
   }
 
   auto node_info = static_cast<ConnextNodeInfo *>(node->data);
-  if (!node_info) {
-    RMW_SET_ERROR_MSG("node info handle is null");
-    return RMW_RET_ERROR;
-  }
+  assert(node_info != nullptr);
```
ivanpauno, line 489: same

rmw_connext_shared_cpp/src/node.cpp
hidmic: Same as in https://github.com/ros2/rmw_fastrtps/pull/408#discussion_r453877085. There is no correct program that may pass a `rmw_node_t` structure that is not fully initialized. If `node_info` or `participant` are `nullptr`, the program is likely corrupt. One may argue though that the likelihood of memory corruption resulting in `nullptr` is low, and so the face value of `assert` vs. not doing anything is equally low. I'd still keep this boy.

hidmic: Which reminds me that we should document that after `rmw_destroy_node()`, using the pointer is UB.

clalancette: > One may argue though that the likelihood of memory corruption resulting in `nullptr` is low, and so the face value of `assert` vs. not doing anything is equally low. I'd still keep this boy.

Right, that's my general argument.  As a library, if it is something that we can reasonably check, then it is unlikely (though not impossible) that it is UB.  If it is UB, then there is nothing that we can do.  So to me, that argues for either doing nothing (there's nothing we can do about UB), or doing a nullptr check-and-return (which would be guarding against a caller of the API accidentally passing a nullptr in, nothing more).

hidmic: > As a library, if it is something that we can reasonably check, then it is unlikely (though not impossible) that it is UB.

I'd reverse that statement. If it is UB, then yes, it's unlikely we can guard ourselves against it. The fact that we can easily check for nullity here doesn't imply anything about it being UB or not. 

> So to me, that argues for either doing nothing (there's nothing we can do about UB), 

True in the most general case. I don't mind dropping the `assert` if you guys think it adds little value.

> doing a nullptr check-and-return (which would be guarding against a caller of the API accidentally passing a nullptr in, nothing more).

This scenario is not equivalent to an accidental `nullptr` argument (for which we do have a check, even though I have mixed feelings about dealing with logical errors in run-time). It can only happen if the `rmw_node_t` structure was intentionally or unintentionally corrupted. Guarding against such bogus code in run-time adds no value, and it only makes debugging harder when the already bad program eventually crashes further down the road.

hidmic:
@clalancette @ivanpauno PTAL https://github.com/ros2/rmw_connext/pull/439/commits/99d15fbf8303e7b1585a198021803106fb98dff3. I had to change the implementation to carry through with the node finalization despite runtime errors. It seemed like it was attempting one of those schemes of successive partial destructions.

rmw_connext_shared_cpp/src/node.cpp
ivanpauno: > There is no correct program that may pass a rmw_node_t structure that is not fully initialized. If node_info or participant are nullptr, the program is likely corrupt. One may argue though that the likelihood of memory corruption resulting in nullptr is low, and so the face value of assert vs. not doing anything is equally low. I'd still keep this boy.

My question is why not keeping the old error checking, I don't see an extra benefit of `assert` over normal error handling.

hidmic: Oof. We just had a rather long discussion with Chris (which I believe he'll post next). The main takeaway is, we shouldn't `assert` nor return an error code. We cannot guard against UB in general (so an `assert` against a particular UB case adds little value), and we shouldn't treat it as an expected or exceptional error condition, it's not (so no return code either, which IMHO is actually harmful in this case as it delays the crash of an already ill-formed or corrupt program).

hidmic:
Running CI up to `test_rmw_implementation` and all RMW implementations:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11444/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6659/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9388/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11347/)

rmw_connext_shared_cpp/src/node.cpp
clalancette: > Oof. We just had a rather long discussion with Chris (which I believe he'll post next)

I posted it over at https://github.com/ros2/rmw_fastrtps/pull/408#discussion_r454341097 , but I forgot we were having essentially the same conversation on two different PRs :).

rmw_connext_shared_cpp/src/node.cpp
```diff
@@ -34,13 +37,42 @@ create_node(
   const char * name,
   const char * namespace_)
 {
-  RCUTILS_CHECK_ARGUMENT_FOR_NULL(context, NULL);
+  assert(implementation_identifier != NULL);
```
clalancette, line 40: This one seems like one where we should `RMW_CHECK_ARGUMENT_FOR_NULL`.  At the very least, that will make it operate the same as the checks for other NULL below, like `context`.

Also, here and elsewhere, we can use `nullptr` instead of `NULL`.

rmw_connext_shared_cpp/src/node.cpp
```diff
@@ -436,74 +468,64 @@ create_node(
 rmw_ret_t
 destroy_node(const char * implementation_identifier, rmw_node_t * node)
 {
-  if (!node) {
-    RMW_SET_ERROR_MSG("node handle is null");
-    return RMW_RET_ERROR;
-  }
+  assert(implementation_identifier != NULL);
```
clalancette, line 471: Same comment about checking and returning on NULL, and using `nullptr` instead.

rmw_connext_shared_cpp/src/node.cpp
hidmic: > This one seems like one where we should RMW_CHECK_ARGUMENT_FOR_NULL. At the very least, that will make it operate the same as the checks for other NULL below, like context.

We could, but given that this is a somewhat internal API (not a standalone RMW implementation) I'd rather not. But I give you that an `assert` here is unwarranted and almost redundant with the check below. Removed it.

rmw_connext_shared_cpp/src/node.cpp
hidmic: I thought about it, but this `rmw_connext*` packages are already inconsistent in this regard pretty much everywhere. If you agree, I'd prefer doing that in a separate PR.

rmw_connext_shared_cpp/src/node.cpp
ivanpauno: > The main takeaway is, we shouldn't assert nor return an error code. We cannot guard against UB in general

I'm ok with the change.
I understand that we cannot guard against general UB, e.g.: the user corrupting a structure with an invalid pointer.

If I'm going to dereference a pointer, I prefer checking that it's not NULL first in some way, but I don't mind strongly about it.

hidmic: > If I'm going to dereference a pointer, I prefer checking that it's not NULL first in some way, but I don't mind strongly about it.

Yeah, that's the usual reflex. And why I initially added an `assert`. But there's nothing special about a `NULL` pointer, besides the fact that it is guaranteed to always be invalid i.e. `ptr != NULL` does not imply `ptr` is valid. I'd assert at most, but otherwise do nothing.

PR #442, opened 2020-07-15
Fix topic creation race condition

ivanpauno:
We were checking if the topic was already creating and then creating it if not.
There's a TOCTTOU race condition here.

To avoid the problem, I've added a `mutex` for each participant, and locking it when creating a topic.

The race condition could be experienced when launching gazebo:
```bash
ros2 launch gazebo_ros gzserver.launch.py
```
though it was extremely hard to reproduce (I was never able to reproduce it, but @jacobperron was able to).

The failure looked like:
- publisher failing
```
    [gzserver-1] terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
    [gzserver-1]   what():  could not create publisher: failed to create topic '/clock' for node namespace='/' name='gazebo', at /home/jacob/ws/ros/gazebo_ros/src/ros2/rmw_connext/rmw_connext_cpp/src/rmw_publisher.cpp:208, at /home/jacob/ws/ros/gazebo_ros/src/ros2/rcl/rcl/src/rcl/publisher.c:172
    [gzserver-1] Aborted (core dumped)
```
- subscription failing
```
[gzserver-1] terminate called after throwing an instance of 'rclcpp::exceptions::RCLError'
[gzserver-1]   what():  could not create subscription: failed to create topic '/clock' for node namespace='/' name='gazebo', at /home/jacob/ws/ros/gazebo_ros/src/ros2/rmw_connext/rmw_connext_cpp/src/rmw_subscription.cpp:202, at /home/jacob/ws/ros/gazebo_ros/src/ros2/rcl/rcl/src/rcl/subscription.c:168
```

I haven't confirmed that gazebo is trying to create the publisher and subscription from different threads, but it sounds like that was the case.

### DIFF
--- a/rmw_connext_cpp/src/rmw_publisher.cpp
+++ b/rmw_connext_cpp/src/rmw_publisher.cpp
@@ -193,6 +193,10 @@ rmw_create_publisher(
   }
 
   topic = rmw_connext_shared_cpp::create_topic(node, topic_name, topic_str, type_name.c_str());
+  if (!topic) {
+    // error already set
+    goto fail;
+  }
   DDS::String_free(topic_str);
   topic_str = nullptr;
 
--- a/rmw_connext_cpp/src/rmw_publisher.cpp
+++ b/rmw_connext_cpp/src/rmw_publisher.cpp
@@ -20,6 +20,7 @@
 #include "rmw/rmw.h"
 #include "rmw/types.h"
 
+#include "rmw_connext_shared_cpp/create_topic.hpp"
 #include "rmw_connext_shared_cpp/qos.hpp"
 #include "rmw_connext_shared_cpp/types.hpp"
 
@@ -122,7 +123,6 @@ rmw_create_publisher(
   DDS::Publisher * dds_publisher = nullptr;
   DDS::DataWriter * topic_writer = nullptr;
   DDS::Topic * topic = nullptr;
-  DDS::TopicDescription * topic_description = nullptr;
   void * info_buf = nullptr;
   void * listener_buf = nullptr;
   ConnextPublisherListener * publisher_listener = nullptr;
@@ -192,34 +192,7 @@ rmw_create_publisher(
     goto fail;
   }
 
-  topic_description = participant->lookup_topicdescription(topic_str);
-  if (!topic_description) {
-    DDS::TopicQos default_topic_qos;
-    status = participant->get_default_topic_qos(default_topic_qos);
-    if (status != DDS::RETCODE_OK) {
-      RMW_SET_ERROR_MSG("failed to get default topic qos");
-      goto fail;
-    }
-
-    topic = participant->create_topic(
-      topic_str, type_name.c_str(),
-      default_topic_qos, NULL, DDS::STATUS_MASK_NONE);
-    if (!topic) {
-      RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
-        "failed to create topic '%s' for node namespace='%s' name='%s'",
-        topic_name, node->namespace_, node->name);
-      goto fail;
-    }
-  } else {
-    DDS::Duration_t timeout = DDS::Duration_t::from_seconds(0);
-    topic = participant->find_topic(topic_str, timeout);
-    if (!topic) {
-      RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
-        "failed to find topic '%s' for node namespace='%s' name='%s'",
-        topic_name, node->namespace_, node->name);
-      goto fail;
-    }
-  }
+  topic = rmw_connext_shared_cpp::create_topic(node, topic_name, topic_str, type_name.c_str());
   DDS::String_free(topic_str);
   topic_str = nullptr;
 
--- a/rmw_connext_cpp/src/rmw_subscription.cpp
+++ b/rmw_connext_cpp/src/rmw_subscription.cpp
@@ -187,6 +187,10 @@ rmw_create_subscription(
   }
 
   topic = rmw_connext_shared_cpp::create_topic(node, topic_name, topic_str, type_name.c_str());
+  if (!topic) {
+    // error already set
+    goto fail;
+  }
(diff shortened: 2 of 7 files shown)


ivanpauno:
I was curious if there was a similar race condition for services and clients.
That code lives [here](https://github.com/ros2/rosidl_typesupport_connext/blob/fbaa793ad686395cd4ee3751f74b308c09faee6e/rosidl_typesupport_connext_cpp/resource/srv__type_support.cpp.em#L206) and [here](https://github.com/ros2/rosidl_typesupport_connext/blob/fbaa793ad686395cd4ee3751f74b308c09faee6e/rosidl_typesupport_connext_cpp/resource/srv__type_support.cpp.em#L105) and there doesn't seem to be any race condition in that case :smiley: .

ivanpauno:
@ros-pull-request-builder retest this please

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11477/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9416/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11374/)

rmw_connext_cpp/src/rmw_publisher.cpp
```diff
@@ -192,34 +192,7 @@ rmw_create_publisher(
...
-    DDS::Duration_t timeout = DDS::Duration_t::from_seconds(0);
-    topic = participant->find_topic(topic_str, timeout);
-    if (!topic) {
-      RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
-        "failed to find topic '%s' for node namespace='%s' name='%s'",
-        topic_name, node->namespace_, node->name);
-      goto fail;
-    }
-  }
+  topic = rmw_connext_shared_cpp::create_topic(node, topic_name, topic_str, type_name.c_str());
```
jacobperron, line 195: I think we need to check `topic == nullptr` and `goto fail` if true.

rmw_connext_cpp/src/rmw_subscription.cpp
```diff
@@ -186,34 +186,7 @@ rmw_create_subscription(
...
-    DDS::Duration_t timeout = DDS::Duration_t::from_seconds(0);
-    topic = participant->find_topic(topic_str, timeout);
-    if (!topic) {
-      RMW_SET_ERROR_MSG_WITH_FORMAT_STRING(
-        "failed to find topic '%s' for node namespace='%s' name='%s'",
-        topic_name, node->namespace_, node->name);
-      goto fail;
-    }
-  }
+  topic = rmw_connext_shared_cpp::create_topic(node, topic_name, topic_str, type_name.c_str());
```
jacobperron, line 189: Same here; check for failure.

rmw_connext_shared_cpp/src/create_topic.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include "rmw_connext_shared_cpp/ndds_include.hpp"
+
+DDS::Topic *
+rmw_connext_shared_cpp::create_topic(
+  const rmw_node_t * node,
+  const char * topic_name,
+  const char * dds_topic_name,
+  const char * dds_topic_type)
+{
+  // This function is internal, and should be called from functions that already verified
```
jacobperron, line 33: Should we consider moving the implementation to a private header?

clalancette: It's a bit subtle, but we shouldn't.  This function is in `rmw_connext_shared_cpp` package, but is used by the `rmw_connext_cpp` package.  So it needs to be in a public header to reach across the package boundary.

There's a related question about whether we should move this function into the `rmw_connext_cpp` package, but see my overall review comment for a reason why not.

rmw_connext_cpp/src/rmw_publisher.cpp
ivanpauno: Done in https://github.com/ros2/rmw_connext/pull/442/commits/a55098b4b4c1d9582cf3e6e1830d7ad26c5c2977

rmw_connext_cpp/src/rmw_subscription.cpp
ivanpauno: Done https://github.com/ros2/rmw_connext/pull/442/commits/a55098b4b4c1d9582cf3e6e1830d7ad26c5c2977

rmw_connext_shared_cpp/src/create_topic.cpp
ivanpauno: I've documented the preconditions for further clarity https://github.com/ros2/rmw_connext/pull/442/commits/a0c3fe1931b984bf259be58600c4b25217711b15.

ivanpauno:
> I think we need this same fix in the dynamic version of the code as well:

Done in https://github.com/ros2/rmw_connext/pull/442/commits/79aa9451d5eeae3c8f41a593ca7c6e1048724828.

I really don't think we should continue updating rmw_connext_dynamic, it is extremely outdated.

IMO, if we want to support it again, it will be easier to rewrite than update it (the only piece of useful code there is how to create a DDS type code from a ROS typesupport and how to convert between ROS and DDS messages with the introspection typesupport).

clalancette:
> I really don't think we should continue updating rmw_connext_dynamic, it is extremely outdated.
> 
> IMO, if we want to support it again, it will be easier to rewrite than update it (the only piece of useful code there is how to create a DDS type code from a ROS typesupport and how to convert between ROS and DDS messages with the introspection typesupport).

Yeah, I basically agree here.  Personally I'd just remove the code from the repository altogether.  If anyone wants to revive it, it would be in git history to serve as a starting point.  I'll bring it up at the next meeting.

jacobperron:
This change (and the follow-up fix in #444) are not ABI compatible due to changes to publicly accessible structs. I'm not sure of a nice way to make them backwards compatible so I'm removing from the Foxy board.

PR #443, opened 2020-07-17
Set context actual domain id

ivanpauno:
See https://github.com/ros2/rcl/pull/718.

### DIFF
--- a/rmw_connext_cpp/src/rmw_init.cpp
+++ b/rmw_connext_cpp/src/rmw_init.cpp
@@ -139,6 +139,8 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
 
   context->instance_id = options->instance_id;
   context->implementation_identifier = rti_connext_identifier;
+  context->actual_domain_id =
+    RMW_DEFAULT_DOMAIN_ID != options->domain_id ? options->domain_id : 0u;
 
   context->impl = new (std::nothrow) rmw_context_impl_t();
   if (nullptr == context->impl) {
--- a/rmw_connext_cpp/src/rmw_init.cpp
+++ b/rmw_connext_cpp/src/rmw_init.cpp
@@ -132,10 +132,8 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
     return RMW_RET_INVALID_ARGUMENT;
   }
 
-  const rmw_context_t zero_context = rmw_get_zero_initialized_context();
-  assert(0 == std::memcmp(context, &zero_context, sizeof(rmw_context_t)));
   auto restore_context = rcpputils::make_scope_exit(
-    [context, &zero_context]() {*context = zero_context;});
+    [context]() {*context = rmw_get_zero_initialized_context();});
 
   context->instance_id = options->instance_id;
   context->implementation_identifier = rti_connext_identifier;
--- a/rmw_connext_cpp/src/rmw_init.cpp
+++ b/rmw_connext_cpp/src/rmw_init.cpp
@@ -139,6 +139,7 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
 
   context->instance_id = options->instance_id;
   context->implementation_identifier = rti_connext_identifier;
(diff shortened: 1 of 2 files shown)


rmw_connext_cpp/src/rmw_init.cpp
```diff
@@ -139,6 +139,8 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
 
   context->instance_id = options->instance_id;
   context->implementation_identifier = rti_connext_identifier;
+  context->actual_domain_id =
+    RMW_DEFAULT_DOMAIN_ID != options->domain_id ? options->domain_id : 0u;
```
hidmic, line 143: @ivanpauno [this line](https://github.com/ros2/rmw_connext/blob/da982d173cdc4d9a4f7ee00683bfefaf743816fd/rmw_connext_shared_cpp/src/node.cpp#L313) has to change too.

ivanpauno: See https://github.com/ros2/rmw_connext/pull/443/commits/c6ac3bdd4103d73ab8ac8153a30cd8a3ca9d2875.

rmw_connext_cpp/src/rmw_init.cpp
```diff
@@ -132,13 +132,14 @@ rmw_init(const rmw_init_options_t * options, rmw_context_t * context)
     return RMW_RET_INVALID_ARGUMENT;
   }
 
-  const rmw_context_t zero_context = rmw_get_zero_initialized_context();
-  assert(0 == std::memcmp(context, &zero_context, sizeof(rmw_context_t)));
```
hidmic, line 136: @ivanpauno unrelated? Also, why dropping this? I recall [the discussion](https://github.com/ros2/rmw_connext/pull/432#discussion_r446426627) that originated it.

ivanpauno: Good question, I should have clarified that.

1. I have added in https://github.com/ros2/rmw/pull/251 [this line](https://github.com/ros2/rmw/blob/3956c762373ab42d22d4c57937db9125c6272b27/rmw/src/init.c#L32). The zero initialized init options aren't actually zero, and thus the comparison fails.
2. I overlooked in #432 this line, that `assert` doesn't make sense. It's checking that the return value of `rmw_get_zero_initialized_context` is full of zeros. If we want to check that, we should write a test in `rmw`.

hidmic: > 1. I have added in ros2/rmw#251 this line. The zero initialized init options aren't actually zero, and thus the comparison fails.

Uh, interesting. We need to get naming straight at some point.

> I overlooked in #432 this line, that assert doesn't make sense. It's checking that the return value of rmw_get_zero_initialized_context is full of zeros. If we want to check that, we should write a test in rmw.

Hmm, it's checking that the given context is zero initialized.

ivanpauno: > Hmm, it's checking that the given context is zero initialized (which is the precondition).

Sorry, I misread the code before :open_mouth: .
I will double check what's going on.

ivanpauno: From https://en.cppreference.com/w/cpp/string/byte/memcmp:

> This function reads object representations, not the object values, and is typically only meaningful for trivially-copyable objects with no padding. For example, memcmp() between two objects of type std::string or std::vector will not compare their contents, and memcmp() between two objects of type struct{char c; int n;} will compare the padding bytes whose values may differ when the values of c and n are the same.

The problem is that the structs are equal, but `memcmp` does still fail.
A random change in `rmw_context_t` layout triggered the problem (adding `actual_domain_id`).
I highly recommend deleting this assert.

In the future, we can add a `rmw_context_compare` or a `rmw_context_is_zero_initialized` function.

hidmic: Argh, didn't think about padding. Ok, let's remove it.

PR #444, opened 2020-07-24
Avoid leaking DDS::Topic objects

ivanpauno:
Follow up of https://github.com/ros2/rmw_connext/pull/442.

Based on [find_topic](https://community.rti.com/static/documentation/connext-dds/5.3.1/doc/api/connext_dds/api_cpp/classDDSDomainParticipant.html#a4495ca19a7c63f82c21a4f07c8200cd9) documentation, it returns a object that can be independently destroyed and not a reference to the one found.

This PR modifies publisher/subscription creation and destruction, so that the `DDS::Topic` objects are destroyed as soon as possible (i.e. when destroying a publisher/subscription).

For example, an application that creates and destroy a publisher on the same topic repeatedly was increasingly allocating more resources. This PR avoids that.

I've also simplified topic creation code, as using both `find_topic` and [lookup_topic_descritpion](https://community.rti.com/static/documentation/connext-dds/5.3.1/doc/api/connext_dds/api_cpp/classDDSDomainParticipant.html#a33b285799bcc1f5383ff990f3c43a1db) is redundant.

This passes `rcl` tests correctly locally, but it deserves full CI before being merged.

---

Publisher and Subscriptions are still leaking the "registered type", but that's much more complex to fix because we're using undocumented rti functions :confused:.

### DIFF
--- a/rmw_connext_shared_cpp/src/create_topic.cpp
+++ b/rmw_connext_shared_cpp/src/create_topic.cpp
@@ -40,6 +40,7 @@ rmw_connext_shared_cpp::create_topic(
 
   std::lock_guard<std::mutex> guard(node_info->topic_creation_mutex);
 
+  // This functions returns a copy, that has to be destroyed independently.
   DDS::Topic * topic = participant->find_topic(dds_topic_name, DDS::Duration_t::from_seconds(0));
   if (!topic) {
     DDS::TopicQos default_topic_qos;
--- a/rmw_connext_shared_cpp/src/create_topic.cpp
+++ b/rmw_connext_shared_cpp/src/create_topic.cpp
@@ -40,8 +40,7 @@ rmw_connext_shared_cpp::create_topic(
 
   std::lock_guard<std::mutex> guard(node_info->topic_creation_mutex);
 
-  DDS::Duration_t timeout = DDS::Duration_t::from_seconds(0);
-  DDS::Topic * topic = participant->find_topic(dds_topic_name, timeout);
+  DDS::Topic * topic = participant->find_topic(dds_topic_name, DDS::Duration_t::from_seconds(0));
   if (!topic) {
     DDS::TopicQos default_topic_qos;
     DDS::ReturnCode_t status = participant->get_default_topic_qos(default_topic_qos);
--- a/rmw_connext_shared_cpp/src/create_topic.cpp
+++ b/rmw_connext_shared_cpp/src/create_topic.cpp
@@ -40,7 +40,7 @@ rmw_connext_shared_cpp::create_topic(
 
   std::lock_guard<std::mutex> guard(node_info->topic_creation_mutex);
 
-  // This functions returns a copy, that has to be destroyed independently.
+  // This function returns a copy that has to be destroyed independently.
   DDS::Topic * topic = participant->find_topic(dds_topic_name, DDS::Duration_t::from_seconds(0));
   if (!topic) {
     DDS::TopicQos default_topic_qos;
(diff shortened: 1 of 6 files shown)


rmw_connext_shared_cpp/src/create_topic.cpp
```diff
@@ -40,14 +40,11 @@ rmw_connext_shared_cpp::create_topic(
 
   std::lock_guard<std::mutex> guard(node_info->topic_creation_mutex);
 
-  DDS::TopicDescription * topic_description =
-    participant->lookup_topicdescription(dds_topic_name);
-
-  DDS::Topic * topic = nullptr;
-  DDS::ReturnCode_t status = DDS::RETCODE_ERROR;
-  if (!topic_description) {
+  DDS::Duration_t timeout = DDS::Duration_t::from_seconds(0);
```
clalancette, line 43: It isn't clear to me from the Connext API docs; does 0 mean "don't wait at all", or "wait forever"?  I'm presuming the former, but I just want to be sure.

rmw_connext_shared_cpp/src/create_topic.cpp
```diff
@@ -40,14 +40,11 @@ rmw_connext_shared_cpp::create_topic(
...
+  DDS::Topic * topic = participant->find_topic(dds_topic_name, timeout);
```
clalancette, line 44: I'm going to suggest that you put a comment above this with a link to the Connext documentation, and a small note saying that we are intentionally getting a copy of the `DDS::Topic` for each thread.  Then it is a little clearer in the future why we are always freeing during destruction.

Separate nit: Since the timeout we are going to use is fixed, you could combine this line and the previous one:

```
DDS::Topic * topic = participant->find_topic(dds_topic_name, DDS::Duration_t::from_seconds(0));
```

But I'll leave it to you what you find more readable.

rmw_connext_shared_cpp/src/create_topic.cpp
ivanpauno: It means "don't wait at all".

rmw_connext_shared_cpp/src/create_topic.cpp
ivanpauno: Done in https://github.com/ros2/rmw_connext/pull/444/commits/7a23fd09fb76bd7a01682a2100f23c2971ad20f1

ivanpauno: I forgot adding the comment, see https://github.com/ros2/rmw_connext/pull/444/commits/3c92dcfc0d6069d496d9d34bf76bc77831169d19.

ivanpauno:
> It's a little unfortunate for performance that we are causing copies of the DDSTopic to be made for every caller that comes in here, but I think this is simpler than doing a reference count or anything like that

Based on rti documentation, it does seem like they are doing reference counting internally and that the copy should be cheap.

rmw_connext_shared_cpp/src/create_topic.cpp
```diff
@@ -40,14 +40,11 @@ rmw_connext_shared_cpp::create_topic(
 
   std::lock_guard<std::mutex> guard(node_info->topic_creation_mutex);
 
-  DDS::TopicDescription * topic_description =
-    participant->lookup_topicdescription(dds_topic_name);
-
-  DDS::Topic * topic = nullptr;
-  DDS::ReturnCode_t status = DDS::RETCODE_ERROR;
-  if (!topic_description) {
+  // This functions returns a copy, that has to be destroyed independently.
```
clalancette, line 43:
```suggestion
  // This function returns a copy that has to be destroyed independently.
```

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11625/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9510/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11600/)

PR #446, opened 2020-07-30
2.3.0

hidmic:
I'm opening this for debate.

I think this is a minor bump because, even though #444 changed some public structs' binary layouts, those structs are not actually part of the package's API (there's actually no reason for them to be in publicly installed headers).

### DIFF
--- a/rmw_connext_cpp/CHANGELOG.rst
+++ b/rmw_connext_cpp/CHANGELOG.rst
@@ -2,8 +2,8 @@
 Changelog for package rmw_connext_cpp
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Forthcoming
------------
+2.3.0 (2020-07-30)
+------------------
 * Ensure compliant publisher API (`#445 <https://github.com/ros2/rmw_connext/issues/445>`_)
 * Avoid leaking DDS::Topic objects (`#444 <https://github.com/ros2/rmw_connext/issues/444>`_)
 * Contributors: Ivan Santiago Paunovic, Michel Hidalgo
--- a/rmw_connext_cpp/CHANGELOG.rst
+++ b/rmw_connext_cpp/CHANGELOG.rst
@@ -2,6 +2,12 @@
 Changelog for package rmw_connext_cpp
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+Forthcoming
+-----------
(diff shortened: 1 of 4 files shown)


hidmic:
@ivanpauno @nuclearsandwich thoughts?

hidmic:
Nevermind the DCO check. If we agree, I'll let `catkin_prepare_release` push to `master`.

nuclearsandwich:
>  public structs' binary layouts, those structs are not actually part of the package's API (there's actually no reason for them to be in publicly installed headers).

IMO if it changes publicly visible structs that's an API change regardless of whether those structs are "meant" to be publicly consumed. But if I'm overridden by other reviewers on this stance I'll yield.

hidmic:
> regardless of whether those structs are "meant" to be publicly consumed

Yes, strictly speaking that is true. But let me further state that those `structs` are not only not meant to be used, they cannot be used (in any meaningful way). No public API takes or returns them, not even as an opaque object. So the only reason that'd justify a major bump is their presence in public headers.

Which makes me wonder. Perhaps we should major bump and make them internal in the process.

nuclearsandwich:
> Which makes me think. Perhaps we should major bump and make them internal in the process. WDYT?

That strictest approach (and reducing unnecessary API surface) is what I favor, but again I'm willing to yield to other opinions if presented.

hidmic:
@dirk-thomas @wjwwood thoughts?

dirk-thomas:
> Perhaps we should major bump and make them internal in the process.

This sounds like the best option to me.

wjwwood:
If the public API is declared (which semver recommends and our QD requires), and the symbol is not in the public API, then I don't think you need a major bump. However, breaking ABI doesn't matter sometimes if it is "public" or not.

If the concern is ABI stability, don't change it. If breaking downstream things due to an API break is the concern, then being in the public API or not matters.

See semver:

> Software using Semantic Versioning MUST declare a public API. This API could be declared in the code itself or exist strictly in documentation. However it is done, it SHOULD be precise and comprehensive.
> ...
> Version 1.0.0 defines the public API. The way in which the version number is incremented after this release is dependent on this public API and how it changes.
> ...
> Major version X (X.y.z | X > 0) MUST be incremented if any backwards incompatible changes are introduced to the **public** API.

(emphasis mine)
https://semver.org/#spec-item-1

hidmic:
Alright, we're going with major bump + making those headers internal (see #447). Thanks for the feedback guys!

PR #451, opened 2020-08-20
Ensure compliant matched pub/sub count API.

hidmic:
Connected to https://github.com/ros2/rmw/pull/262.

### DIFF
--- a/rmw_connext_cpp/src/rmw_publisher.cpp
+++ b/rmw_connext_cpp/src/rmw_publisher.cpp
@@ -333,25 +333,15 @@ rmw_publisher_count_matched_subscriptions(
   const rmw_publisher_t * publisher,
   size_t * subscription_count)
 {
-  if (!publisher) {
-    RMW_SET_ERROR_MSG("publisher handle is null");
-    return RMW_RET_INVALID_ARGUMENT;
-  }
-
-  if (!subscription_count) {
-    RMW_SET_ERROR_MSG("subscription_count is null");
-    return RMW_RET_INVALID_ARGUMENT;
-  }
+  RMW_CHECK_ARGUMENT_FOR_NULL(publisher, RMW_RET_INVALID_ARGUMENT);
+  RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
+    publisher,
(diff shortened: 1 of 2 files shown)


rmw_connext_cpp/src/rmw_publisher.cpp
```diff
@@ -333,25 +333,15 @@ rmw_publisher_count_matched_subscriptions(
...
 
   auto info = static_cast<ConnextStaticPublisherInfo *>(publisher->data);
-  if (!info) {
-    RMW_SET_ERROR_MSG("publisher internal data is invalid");
-    return RMW_RET_ERROR;
-  }
-  if (!info->listener_) {
-    RMW_SET_ERROR_MSG("publisher internal listener is invalid");
-    return RMW_RET_ERROR;
-  }
```
clalancette, line 354: Why are we removing these checks?

(sorry if this is redundant with what we have talked about before)

hidmic: See https://github.com/ros2/rmw_fastrtps/pull/408#discussion_r454341097. I can't find the discussion about not duplicating checks across `rmw_fastrtps_shared_cpp` and the actual Fast-RTPS based RMW implementation.

But long story short:
- There's no point in checking arguments for nullity twice -- once in `rmw_fastrtps(_dynamic)_cpp`, once in `rmw_fastrtps_shared_cpp`.
- There's no point in guarding ourselves those particular UB cases. If an `rmw_fastrtps(_dynamic)_cpp` publisher (or subscription) has `nullptr` internals, either the user is using the structure in an unintended way (and the program is logically broken) or memory corruption ensued.

hidmic:
CI up to `test_rmw_implementation`, against all Tier 1 RMW implementations:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12070/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7095/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9844/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12001/)

PR #458, opened 2020-09-18
Fix wrong error messages

ivanpauno:
Minor errors found while doing another thing.


jacobperron (approved):

jacobperron:
@ivanpauno If you have time to backport this that would be great :)

PR #460, opened 2020-09-21
Allow more flexible usage of XML QoS profiles

ivanpauno:
It allows more flexibility through environment variables, checkout `README.md` for more details.

~TODO: Write tests.~

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ export RMW_CONNEXT_QOS_PROFILE_LIBRARY=Ros2TestQosLibrary
 
 You can use the `RMW_CONNEXT_DEFAULT_QOS_PROFILE` environment variable for this.
 It overrides the profile marked with `is_default_qos="true"` when set.
-The profile is looked up in the qos profile library RMW connext is using.
+The profile is looked up in the QoS profile library RMW connext is using.
 
 ### Using user provided publish mode
 
--- a/README.md
+++ b/README.md
@@ -29,7 +29,8 @@ Refer to the [Install DDS Implementations](https://index.ros.org/doc/ros2/Instal
 QoS profiles can be specified in XML according to the load order specified [here](https://community.rti.com/static/documentation/connext-dds/5.2.0/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/How_to_Load_XML_Specified_QoS_Settings.htm). `url_profile` and `string_profile` cannot be used.
 
 ROS will use the profile with the `is_default_qos="true"` attribute.
-That profile will be used as a base and it will be overridden with the ROS specific policies except when `rmw_qos_profile_system_default` is used.
+That profile will be used as a base and it will be policies defined in the ROS qos profile will be overriden, except when `rmw_qos_profile_system_default` is used.
+
 For example:
 
 ```xml
@@ -67,6 +68,9 @@ xsi:noNamespaceSchemaLocation="http://community.rti.com/schema/5.3.1/rti_dds_qos
       </participant_qos>
 
       <datawriter_qos topic_filter="rt/my_large_data_topic">
+        <reliability>
+          <kind>RELIABLE_RELIABILITY_QOS</kind>
+        </reliability>
         <publish_mode>
           <flow_controller_name>dds.flow_controller.token_bucket.slow_flow</flow_controller_name>
         </publish_mode>
@@ -76,23 +80,26 @@ xsi:noNamespaceSchemaLocation="http://community.rti.com/schema/5.3.1/rti_dds_qos
 </dds>
 ```
 
-That will force all publishers in the `my_large_data_topic` to use the `slow_flow` flow controller.
+That will force all publishers in the `my_large_data_topic` to use the `slow_flow` flow controller, but the reliability specified in the ROS QoS profile will be used except if its value is `RMW_QOS_RELIABILITY_POLICY_SYSTEM_DEFAULT`. 
 See `Topic Name Mangling` section to understand the `rt/` prefix.
 
-### Using different profiles for different nodes
+See [RTI Connext docs](https://community.rti.com/static/documentation/connext-dds/5.2.0/doc/manuals/connext_dds/html_files/RTI_ConnextDDS_CoreLibraries_UsersManual/Content/UsersManual/Topic_Filters.htm) to understand topic filters.
+
+### Overriding ROS specified QoS policies for a topic
 
 To use this feature, you must first set the following environment variable:
 
 ```bat
 :: Windows
-set RMW_CONNEXT_ALLOW_NODE_QOS_PROFILES=1
+set RMW_CONNEXT_ALLOW_TOPIC_QOS_PROFILES=1
 ```
 ```bash
 # Linux/MacOS
-export RMW_CONNEXT_ALLOW_NODE_QOS_PROFILES=1
+export RMW_CONNEXT_ALLOW_TOPIC_QOS_PROFILES=1
 ```
 
-If the environment variable is set, if a profile name matches the fully qualified node name (e.g.: `/my/full/namespace/my_node_name`), it will be used instead of the default profile.
+If the environment variable is set, when a profile name matches the dds topic name, it will be used and the ROS specified profile will be ignored.
+
 For example:
 ```xml
 <?xml version="1.0"?>
@@ -128,24 +135,29 @@ xsi:noNamespaceSchemaLocation="http://community.rti.com/schema/5.3.1/rti_dds_qos
         </property>
       </participant_qos>
     </qos_profile>
-    <qos_profile name="/my_node" base_name="BuiltinQosLib::Baseline.5.3.0">
-      <datawriter_qos topic_filter="rt/my_large_data_topic">
+    <qos_profile name="rt/my_large_data_topic" base_name="BuiltinQosLib::Baseline.5.3.0">
+      <datawriter_qos>  <!--Don't use topic filters here-->
         <publish_mode>
           <flow_controller_name>dds.flow_controller.token_bucket.slow_flow</flow_controller_name>
         </publish_mode>
+        <reliability>
+          <kind>RELIABLE_RELIABILITY_QOS</kind>
+        </reliability>
       </datawriter_qos>
     </qos_profile>
   </qos_library>
 </dds>
 ```
 
-Will use the `slow_flow` flow controller for publishers in the `my_large_data_topic` of the `/my_node` node.
-All other nodes will use the default profile.
+In this case, all publishers in the topic `/my_large_data_topic` will use the specified slow flow controller and have a reliable reliability (regardless of the reliability specified in ROS code).
 
-The profiles matching the node name will only be used for `datawriter` and `datareader` qos policies.
-Participant/Publisher/Subscription policies will always use the default profile.
+Caveats:
 
-RTI Connext will log an error each time that it tries to find a profile that doesn't exist, if you didn't add a profile for each node those logs will appear.
+- If you want to override the QoS profiles used for all publishers in a topic, the subscription profiles in the same topic will also be overriden. If you don't explicitly provide one, a default will be used.
+- RTI Connext will log an error each time that it tries to find a profile that doesn't exist.
+  Your will see a lot of these logs in your terminal when using `RMW_CONNEXT_ALLOW_TOPIC_QOS_PROFILES` option. 
+
+### Specifying an specific QoS library
 
 If you only provided one QoS library to the process, that one will be used.
 If not, the `RMW_CONNEXT_QOS_PROFILE_LIBRARY` must be used:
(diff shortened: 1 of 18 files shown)


ivanpauno:
@jacobperron I was originally thinking to add this in a fork, but if the defaults aren't modified and flexibility is provided through environment variables, I guess we can merge the change on master.

ivanpauno:
BTW, this is ready for being reviewed. I opened it as a draft only because tests are missing.

rmw_connext_shared_cpp/src/qos.cpp
```diff
@@ -156,34 +162,81 @@ bool
...
+      RMW_SET_ERROR_MSG("failed to get participant factory");
+      return false;
+    }
+    char fqnn[RMW_NAMESPACE_MAX_LENGTH + RMW_NODE_NAME_MAX_NAME_LENGTH];
+    if ('/' == namespace_[0] && '\0' == namespace_[1]) {
+      // root namespace
+      std::snprintf(fqnn, size_t(fqnn), "/%s", node_name);
+    } else {
+      // non-root namespace
+      std::snprintf(fqnn, size_t(fqnn), "%s/%s", namespace_, node_name);
```
jacobperron, line 183: Based on the if-condition, `namespace_` doesn't necessarily have a leading forward slash `/`. Should we handle this case?

rmw_connext_shared_cpp/src/qos.cpp
```diff
@@ -156,34 +162,81 @@ bool
...
-    RMW_SET_ERROR_MSG("failed to get default datareader qos");
-    return false;
+  bool profile_found = false;
+  if (are_node_profiles_allowed()) {
+    DDS::DomainParticipantFactory * dpf = DDS::DomainParticipantFactory::get_instance();
+    if (!dpf) {
+      RMW_SET_ERROR_MSG("failed to get participant factory");
+      return false;
+    }
+    char fqnn[RMW_NAMESPACE_MAX_LENGTH + RMW_NODE_NAME_MAX_NAME_LENGTH];
```
jacobperron, line 177: Account for null character?
```suggestion
    char fqnn[RMW_NAMESPACE_MAX_LENGTH + RMW_NODE_NAME_MAX_NAME_LENGTH + 1];
```

rmw_connext_shared_cpp/src/init.cpp
```diff
@@ -12,23 +12,242 @@
...
+
+/// Tri-state retcode used in `set_default_qos_library` and `is_env_variable_set`.
+enum class TristateRetCode {SET, NOT_SET, FAILED};
+
+/// Set the default Connext qos profile library.
+/**
+ * If `RMW_CONNEXT_QOS_PROFILE_LIBRARY` environment variable is set,
+ * the library matching that name is used.
+ * If only one user provided library is found, that library is set as the default one.
+ * If not, not qos profile library is set.
```
jacobperron, line 43:
```suggestion
 * If not, no qos profile library is set.
```

rmw_connext_shared_cpp/src/init.cpp
```diff
@@ -12,23 +12,242 @@
...
+is_env_variable_set(const char * env_var_name);
+
 rmw_ret_t
 init()
 {
-  DDS::DomainParticipantFactory * dpf_ = DDS::DomainParticipantFactory::get_instance();
-  if (!dpf_) {
-    RMW_SET_ERROR_MSG("failed to get participant factory");
-    return RMW_RET_ERROR;
+  rmw_ret_t ret = RMW_RET_OK;
```
jacobperron, line 78: Should this be `static`? Otherwise, a second call to `init()` will always return `RMW_RET_OK`, even if the first call returned `RMW_RET_ERROR`.

rmw_connext_shared_cpp/src/init.cpp
```diff
@@ -12,23 +12,242 @@
...
+    for (int i = 0; i < qos_libraries.length(); i++) {
+      if (
+        strcmp("BuiltinQosLib", qos_libraries[i]) != 0 &&
+        strcmp("BuiltinQosLibExp", qos_libraries[i]) != 0)
+      {
+        if (!qos_profile_library_name) {
+          qos_profile_library_name = qos_libraries[i];
+        } else {
+          qos_profile_library_name = NULL;
+          break;
```
jacobperron, line 181: This logic looks odd to me. IIUC, we're only setting the library name if there is exactly one was loaded that is not one of the two built-ins (and there are only ever two built-ins?). Is that right? If so, maybe adding a comment here would help clarify:
```suggestion
          // More than one QoS profile library was loaded, do not set
          qos_profile_library_name = NULL;
          break;
```

rmw_connext_shared_cpp/src/qos.cpp
ivanpauno: > Based on the if-condition, namespace_ doesn't necessarily have a leading forward slash /. Should we handle this case?

Namespaces do always have a leading forward slash: https://github.com/ros2/rmw/blob/8a29332a3b6f8f3de6809426ef22e327c7927589/rmw/include/rmw/validate_namespace.h#L64.

In some places this constrain is relaxed, but not in `rmw`.

rmw_connext_shared_cpp/src/qos.cpp
ivanpauno: Good catch

rmw_connext_shared_cpp/src/init.cpp
ivanpauno: Great catch, fixed in https://github.com/ros2/rmw_connext/pull/460/commits/0cb33be75001cca34a31b5920cb0344bebd5e3c8.

rmw_connext_shared_cpp/src/init.cpp
ivanpauno: > and there are only ever two built-ins?

yeah, https://community.rti.com/examples/built-qos-profiles.

> Is that right? If so, maybe adding a comment here would help clarify:

Sounds good

README.md
```diff
@@ -24,5 +24,169 @@ Note that the RTI website has [Free Trial](https://www.rti.com/free-trial) offer
...
+          <flow_controller_name>dds.flow_controller.token_bucket.slow_flow</flow_controller_name>
+        </publish_mode>
+      </datawriter_qos>
+    </qos_profile>
+  </qos_library>
+</dds>
+```
+
+That will force all publishers in the `my_large_data_topic` to use the `slow_flow` flow controller.
+See `Topic Name Mangling` section to understand the `rt/` prefix.
```
mallanmba, line 80: where do I find the `Topic Name Mangling` section?

README.md
```diff
@@ -24,5 +24,169 @@ Note that the RTI website has [Free Trial](https://www.rti.com/free-trial) offer
...
+set RMW_CONNEXT_QOS_PROFILE_LIBRARY=Ros2TestQosLibrary
+```
+```bash
+# Linux/MacOS
+export RMW_CONNEXT_QOS_PROFILE_LIBRARY=Ros2TestQosLibrary
+```
+
+### Ignoring the ROS specified QoS
+
+If you want to override the QoS profile specified in ROS code, you can do that by setting:
```
mallanmba, line 163: Is setting an environment variable required in order to override a QoS? From my understanding of the text above, the ROS QoS will be used unless I have a profile with an appropriate `topic_filter` attribute, in which case that topic will use the specified QoS without any "opt-in" environment variable being set. 
If `RMW_CONNEXT_IGNORE_ROS_QOS` is set, then all topics will use the default QoS, unless there is a QoS with an appropriate `topic_filter` attribute. 
Is that correct?

README.md
ivanpauno: Sorry, I forgot writing it.

README.md
ivanpauno: > Is setting an environment variable required in order to override a QoS?

It's not needed for setting things like "Flow controllers", which ROS doesn't know about.
But it's needed to override the things included in the ROS QoS profile.

> From my understanding of the text above, the ROS QoS will be used unless I have a profile with an appropriate topic_filter attribute, in which case that topic will use the specified QoS without any "opt-in" environment variable being set.

To be more clear:
```xml
    <qos_profile name="default" is_default_qos="true">
      <datawriter_qos topic_filter="rt/my_topic">
        <reliability>reliable</reliability>
      </datawriter_qos>
    </qos_profile>
```
would not override the reliability QoS for publishers in topic "/my_topic" except:
- ros reliability is set to "system default"
- RMW_CONNEXT_IGNORE_ROS_QOS is set.

> If RMW_CONNEXT_IGNORE_ROS_QOS is set, then all topics will use the default QoS, unless there is a QoS with an appropriate topic_filter attribute.

Yes (and there could be also a profile matching the name of the node).
This might be a bit error prone, because the ROS specified qos profiles are never used.

---

This is maybe a bit different of what I originally said in the meeting.
I diverged a bit because of wanting to use the "topic filters" feature, but maybe that was not a good idea.

The reason why `RMW_CONNEXT_IGNORE_ROS_QOS` is need is because there's no way to check if you're using the default qos or one that matches a topic filter (Connext doesn't provide an API to query that).

If profile names are used instead of the topic filter feature, then it's possible to do the following:
- Use the profile matching the topic name (if there's one).
- Use the ros profile, if there's no profile matching the name.

In that case it's not possible to use different QoS in different nodes for the same topic, but that maybe isn't really useful.
This alternative may be less error prone to the current PR.

There's another issue: though you can query if a profile of a given name exists, if it doesn't exist Connext will always logs an error (which will generate some noise in the logs at startup).
The only way to avoid the logs is to temporally deactivate Connext logging, that might be bad because you can lose an error log from another thread (the logger verbosity is not "thread local").

---

To sum up:
- We can use the approach of this PR, and set `RMW_CONNEXT_IGNORE_ROS_QOS`.
  That requires to carefully specify "all" QoS policies in the file (at least, all the ones that cannot be the default).
- Use the approach here, and set all ROS QoS policies that should be overridden to "system default".
  Because of "topic filters", you can choose different profiles for different topics.
- Don't use topic filters and look for a profile name matching the topic name. That will not allow wildcards, but it won't require setting the ros profile to "system default" in the places where an override is desired and will be less error prone that using the `RMW_CONNEXT_IGNORE_ROS_QOS` environment variable.

@mallanmba which alternative sounds more desirable?
Using profile names instead of topic filters is an easy change.

cc @jacobperron

README.md
ivanpauno: Added in https://github.com/ros2/rmw_connext/pull/460/commits/db3b1d08367ba9f580977850e0631300d8a05630.

README.md
ivanpauno: Updated after yesterday discussion in https://github.com/ros2/rmw_connext/pull/460/commits/1aae68ac33cdd31a5882d16e32753bc87171c844.

The current approach allows you to specify QoS in two ways:
- A profile marked `is_default_true="true"` with topic filters. In that case you cannot override non-system default QoS policies, but specifying things like a flow controller is possible. Doesn't require any environment variable set.
- If a profile matching the DDS topic name (see name mangling) is found, those will override the ROS QoS.
  Requires environment variable `RMW_CONNEXT_ALLOW_TOPIC_QOS_PROFILES=1`.

jacobperron (approved):
I didn't exercise the feature, but it LGTM. One wording suggestion.

ivanpauno:
@jacobperron I'm planning to write some tests before merging, to improve coverage.
We could kick a packaging build with this branch, I have done some manual testing and it's working fine.

ivanpauno:
Tests added, CI (testing rcl, test_rmw_implementation, rmw_connext_cpp, rmw_connext_shared_cpp):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12690/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7653/) (failed because rti Connext isn't available in aarch64 in our CI)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10405/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12634/)

README.md
```diff
@@ -24,5 +24,178 @@ Note that the RTI website has [Free Trial](https://www.rti.com/free-trial) offer
...
+set RMW_CONNEXT_DO_NOT_OVERRIDE_PUBLICATION_MODE=1
+```
+```bash
+# Linux/MacOS
+export RMW_CONNEXT_DO_NOT_OVERRIDE_PUBLICATION_MODE=1
+```
+
+## ROS topic name mangling
+
+ROS uses the following mangled topics when the ROS qos policy `avoid_ros_namespace_conventions` is `false`, which is the default:
```
jacobperron, line 195:
```suggestion
ROS uses the following mangled topics when the ROS QoS policy `avoid_ros_namespace_conventions` is `false`, which is the default:
```

README.md
```diff
@@ -24,5 +24,178 @@ Note that the RTI website has [Free Trial](https://www.rti.com/free-trial) offer
...
+```bash
+# Linux/MacOS
+export RMW_CONNEXT_QOS_PROFILE_LIBRARY=Ros2TestQosLibrary
+```
+
+### Specifying a different default QoS profile
+
+You can use the `RMW_CONNEXT_DEFAULT_QOS_PROFILE` environment variable for this.
+It overrides the profile marked with `is_default_qos="true"` when set.
+The profile is looked up in the qos profile library RMW connext is using.
```
jacobperron, line 177:
```suggestion
The profile is looked up in the QoS profile library RMW connext is using.
```

jacobperron (approved):
LGTM, just some tiny nitpicks

ivanpauno:
Going in, thanks for the reviews @jacobperron !

mjcarroll:
@ivanpauno There is a chance that this introduced flakiness in the nightly Windows builds.  Do you mind taking a look and seeing if it was this PR? These are the tests in question: https://ci.ros2.org/job/nightly_win_rel/1721/testReport/

ivanpauno:
> @ivanpauno There is a chance that this introduced flakiness in the nightly Windows builds. Do you mind taking a look and seeing if it was this PR? These are the tests in question: https://ci.ros2.org/job/nightly_win_rel/1721/testReport/

I don't see how this PR can make that test fail, it's completely unrelated.
That test is failing to find that some nodes were set up within a timeout, and this PR cannot affect node discovery (I would be extremely surprised if it somehow does).

Most likely, the test was already flaky.

mjcarroll:
I didn't think so, but I also hadn't seen these tests pop up in recent history.

I'll see if I can reproduce, thanks!

PR #461, opened 2020-09-21
Update gid API return codes.

hidmic:
To be compliant with the API spec. Connected to https://github.com/ros2/rmw/pull/274.

### DIFF
--- a/rmw_connext_cpp/src/rmw_compare_gid_equals.cpp
+++ b/rmw_connext_cpp/src/rmw_compare_gid_equals.cpp
@@ -25,38 +25,22 @@ extern "C"
 rmw_ret_t
 rmw_compare_gids_equal(const rmw_gid_t * gid1, const rmw_gid_t * gid2, bool * result)
 {
-  if (!gid1) {
-    RMW_SET_ERROR_MSG("gid1 is null");
-    return RMW_RET_ERROR;
-  }
+  RMW_CHECK_ARGUMENT_FOR_NULL(gid1, RMW_RET_INVALID_ARGUMENT);
   RMW_CHECK_TYPE_IDENTIFIERS_MATCH(
     gid1,
     gid1->implementation_identifier,
     rti_connext_identifier,
-    return RMW_RET_ERROR)
-  if (!gid2) {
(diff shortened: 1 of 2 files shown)


hidmic:
CI up to `test_rmw_implementation`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12383/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7352/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10095/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12296/)

rmw_connext_cpp/src/rmw_compare_gid_equals.cpp
```diff
@@ -25,38 +25,22 @@ extern "C"
...
-    return RMW_RET_ERROR;
-  }
+    return RMW_RET_INCORRECT_RMW_IMPLEMENTATION);
+  RMW_CHECK_ARGUMENT_FOR_NULL(result, RMW_RET_INVALID_ARGUMENT);
+
   auto detail1 = reinterpret_cast<const ConnextPublisherGID *>(gid1->data);
-  if (!detail1) {
-    RMW_SET_ERROR_MSG("gid1 is invalid");
-    return RMW_RET_ERROR;
-  }
```
clalancette, line 54: So can it be the case that `gid1->data` is nullptr?  Or are we just taking the approach of elsewhere where nullptr is just a special case of "undefined pointer"?

clalancette (commented):
One question.

Lobotuerk (approved):
LGTM

rmw_connext_cpp/src/rmw_compare_gid_equals.cpp
hidmic: `gid1->data` won't ever be `nullptr` -- [it's a primitive array](https://github.com/ros2/rmw/blob/6d3f68af511bb17dfdc9eef3dc345469e8042162/rmw/include/rmw/types.h#L475).

clalancette (approved):

